PR #20, opened 2021-01-12
Download chirps

Tinkaa:
Basic download and processing functionalities for CHIRPS observational data.
Used as basis for MWI experimentation


joseepoirier (approved):
looks good. aligns with our conversation in slack

PR #21, opened 2021-01-12
FI worldpop refactor

Tinkaa:
Refactor FN processing with worldpop data script for readability and small improvements such as consistent column names. 
Waiting with deleting our other methodology (based on subnational data) till we know why the subnational totals differ from worldpop


joseepoirier (approved):
Good documentation of the problem. It seems that the discrepancies in recent years got smaller.

Should we include at the bottom what decision we made regarding which number to use?

Tinkaa:
> Good documentation of the problem. It seems that the discrepancies in recent years got smaller.
> 
> Should we include at the bottom what decision we made regarding which number to use?

Yes, will do in a next PR :)

PR #22, opened 2021-01-15
FI GlobalIPC aggregate

Tinkaa:
Change the methodology of Global IPC processing on two aspects:
- recompute the population per district (often admin2) by summing the populations of the 5 IPC phases, instead of using the reported population. Done cause we saw discrepancies between the two for some entries. 
- aggregate to admin1 and admin0 by summing the numbers of the districts. The data also report pre-computed numbers on admin1 and admin0 level but they don't always correspond, so for consistency, we want to sum them ourselves. 

Also, code has been implemented to raise a warning if one of these two issues occur. The code will still be run, but it is good to have an indication of the quality of the data.

### DIFF
--- a/analyses/malawi/notebooks/GIPC_erroradm0aggr.ipynb
+++ b/analyses/malawi/notebooks/GIPC_erroradm0aggr.ipynb
@@ -5,7 +5,9 @@
    "metadata": {},
    "source": [
     "## Disreperancy Global IPC aggregation methods\n",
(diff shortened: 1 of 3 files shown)


joseepoirier (approved):
Clear comments. Similarly to the FewsNet scripts, I wonder if we should end the document by stating what solution we decided to use to solve for the discrepancies?

PR #25, opened 2021-01-22
Fi computetrigger 2

Tinkaa:
Note: same as branch fi-computetrigger but due to some weirdness it doesn't let me create a PR from that branch 

- rerun all analysis of countries with updated code of last few PRs
- clarify eth trigger analysis notebook
- add notebook investigating differences in eth with the two methods of aggregation for FN data


utils_general/plotting.py
```diff
@@ -239,4 +239,45 @@ def plot_spatial_columns(df, col_list, title=None, predef_bins=None,cmap='YlOrRd
         fig.suptitle(title, fontsize=14, y=0.92)
     fig.tight_layout()
 
+    return fig
```
joseepoirier, line 242: is there a more descriptive name we can give this file?

joseepoirier (approved):
Thanks for all the refactoring!

PR #29, opened 2021-02-01
Fewsnet dates

Tinkaa:
Add two columns with months for which the short and medium-term projections (ML1 and ML2) are valid. 
Now added as a string, in future can be changed to a pandas date range object if that suits the usage better


Tinkaa:
@joseepoirier I am gonna merge this for now as I want to use it in another branch but feel free to comment later.

PR #32, opened 2021-02-05
Refactor bangladesh

turnerm:
A proposed new structure for the `bangladesh` portion of the PA repo. 

Since the trigger and pilot evaluation do not have significant functionality overlap, I propose to separate their respective scripts into two different directories. If there is anything shared it should probably go into `utils_general` (like the GloFAS script; but that will need some refactoring to be moved). 

I still haven't updated the README because I want to see what you think about it first.

Something that should be specified in the README is where the scripts should be run from, since the hard-coded directory paths are sensitive to this. I've modified the files so that they are all intended to be run from the `bangladesh` directory, let me know what you think. The other option is to intend them to be run them from the `pilot_evaluation` and `trigger_analysis` directories but I think that's slightly less nice because then the `data` and `results` directories are above.


hannahker:
Yes, agree with the separation between pilot evaluation and trigger analysis, as well as running the scripts from the top bangladesh directory. I think the ```notebooks``` could also probably be specific to either the pilot evaluation or the trigger analysis

hannahker (approved):
Think it looks great to be merged, probably with a quick update to the README.

PR #33, opened 2021-02-09
Mwi rainfall setup

joseepoirier:
Added years 2011-2020. Dry spells in those years and that bridge two years should be picked up.

All dry spells are at least 15 days. I expected some to be exactly 14. Can you poke around and see if I'm adding an extra day or overlooking something when calculating first, last day and duration of dry spell?

Lastly, it seems there is no observational data for two districts: Likoma and Mwanza. I haven't look into why yet. Any ideas off the top of your heads? The CHIRPS rasters should cover those areas even if they were part of other districts until recently.


Tinkaa:
Regarding CHIRPS data for Likoma and Mwanze:
Did a QGIS check and CHIRPS data is covering the whole region, including those two districts. But what I think is happening, is that none of the raster cells has its centre within the admin boundary… 
Two solutions could be to
- Use the 5 degree CHIRPS data instead of 25 that we are using now, though processing might be much longer
- Use all cells that touch an area instead of all cells that have their centre inside the area. This has as disadvantage that the raster region you are taking into account might be larger than the actual admin region though

analyses/malawi/scripts/mwi_chirps_dry_spell_detection.R
```diff
@@ -0,0 +1,127 @@
...
+
+## identify dry spells per adm2
+
+# list days on which 14-day rolling sum is 2mm or less of rain
+data_max_sums$rollsum_ds_bin <- ifelse(data_max_sums$rollsum_14d <= 2, 1, 0)
+
+# identify beginning, end and duration of dry spells per region
+dry_spells_list <- data_max_sums %>%
+                      mutate(rollsum_ds_bin = ifelse(is.na(rollsum_ds_bin), 0, rollsum_ds_bin)) %>%  # remove NAs with 0
+                      group_by(pcode, spell = cumsum(c(0, diff(rollsum_ds_bin) != 0))) %>% # groups consecutive days with rolling sum <= 2mm
```
Tinkaa, line 104: Regarding the minimum reported dry spell being 15 days while you would expect there to be also 14 day dryspells: could it be that this line should be `cumsum(c(0, diff(rollsum_ds_bin) **== 1**)))` ? I am not fully understanding the code yet, but in `rollsum_ds_bin` there are also -1's which are now counted in the same way as 1. If I change this line, I do get 14 day period dryspells. And an example where a 14 day period dryspell occurred is  on `last_dry_spell_date=2010-06-30` in MW101

Tinkaa:
General comment: great job Josée!

analyses/malawi/scripts/mwi_chirps_dry_spell_detection.R
joseepoirier: Hm. I only have 1's and 0's under `rollsum_ds_bin`. That variable flag days on which the 14-day trailing sum is less then or equal to 2mm (ie a binary indicating if the rolling sum indicates a dry spell).

Ln104 is admittedly hairy. It's essentially assigning a "streak number" to series of rows that share a `rollsum_ds_bin` value. The series for which `rollsum_ds_bin` == 1 are dry spells since the rolling sum is over 14 days. 

I think I found the bug: Ln105. It should read n() >= 1. In the case you pointed out, there was a single date on which the 14-day sum was <= 2mm (ie `rollsum_ds_bin == 1`). The dry spell was exactly 14 days. Because it's only on its last day that the definition of dry spell was met, it was a single record of a dry spell and that case was filtered out by n() > 1.

joseepoirier: Let me know if this works for you (no -1's under `rollsum_ds_bin`, presence of 14-day spells, and flagging of the streak ending on 2010-06-30 in MW101 as a dry spell. 

Thanks for helping figure this out!

hannahker:
Did an initial run through of the ```mwi_chirps_dry_spell_detection.R``` file, which works fine for me, although I haven't had the chance to look into things too closely. 

@joseepoirier can confirm your points above that in ```data_max_sums``` there are no negative values for ```rollsum_ds_bin```, yes to 14-day spells, and yes to seeing a dry spell ending in 2010-06-30 in MW101. 

Happy to look into the code further if there are still some outstanding questions, although I might need to look up some documentation on CHIRPS data since I'm not very familiar with it.

analyses/malawi/scripts/mwi_chirps_dry_spell_detection.R
Tinkaa: Oops yes, something went wrong with the minusses in me poking the code, but think the n>=1 makes sense

joseepoirier:
@turnerm Per our agreement you can merge this branch into main so the restructuring of the repo can take place. This is time-sensitive work however so I'll need to open a new branch Monday to continue the work.

Tagging @Tinkaa in case there is a concern over merging this branch that I am not seeing now.

Tinkaa:
Added few local changes I still had regarding downloading CHIRPS 5 degree resolution data. Good to merge imo

PR #35, opened 2021-02-17
Eth trigger

Tinkaa:
- Update `ethiopia/scripts/eth_foodinsec_trigger.py` to use fewsnet aggregation based on admin2 population instead of worldpop
- Include Jan 2021 update
- Update dashboard to include new seasonal forecasts, updated shapefile and accept slightly different format of the output of `ethiopia/scripts/eth_foodinsec_trigger.py`

The large number of files changed is due to 1) added shapefiles and 2) changes related to SOM&CHD, so would help to first merge #30


dashboard/generate_subnational_map.R
```diff
@@ -7,71 +7,86 @@ library(sf)
 library(tmap)
 
 # read in data to display
-ipc_indices_data <- read.csv("Nov2020_FewsNetGlobalIPC.csv")
+ipc_indices_data <- read.csv("data/foodinsecurity/ethiopia_foodinsec_trigger.csv") #Nov2020_FewsNetGlobalIPC.csv")
```
joseepoirier, line 10: should this csv now be for all countries? does the new dashboard need a specific format?

dashboard/generate_subnational_map.R
```diff
@@ -7,71 +7,86 @@ library(sf)
...
 eth_fn <-  eth_adm1 %>%
-             left_join(ipc_indices_data[ipc_indices_data$Source == 'FewsNet',], by = c('ADM1_EN' = 'ADMIN1'))
+             left_join(ipc_indices_data_latest_fn, by = c('ADM1_EN' = 'ADMIN1'))
+
+# ETH GLB  
+# For global IPC, it can be the case that for some dates, some admins do have ML1 projections but no ML2.
+# In this case we want to show those admins as missing in the ML2 map
+# Therefore, only select the admins with ML1/ML2 projections (indicated by a not-nan ML1/ML2 population)
+eth_gbl_ml1 <-  eth_adm1 %>%
+              left_join(ipc_indices_data_latest_gbl %>% drop_na(pop_ML1), by = c('ADM1_EN' = 'ADMIN1'))
```
joseepoirier, line 29: why would we want to drop a whole row if it only has one set of projections? And if we do, shouldn't we drop_na(pop_ML2) since these are the projections that are sometimes missing?

joseepoirier: Ok I see you address missing ML2 in the next few lines. Are there cases where we lack ML1s?

dashboard/server.R
```diff
@@ -79,12 +79,12 @@ server <- function(input, output) {
...
     output$chc <- renderText({c('<img src="','https://blog.chc.ucsb.edu/wp-content/uploads/2021/01/Screen-Shot-2021-01-20-at-5.21.51-PM.png"','width = "500px" height = "500px"', '>')})
-    output$nmme <- renderText({c('<img src="','https://www.cpc.ncep.noaa.gov/products/NMME/prob/images/prob_ensemble_prate_season2.png"','width = "500px" height = "500px"', '>')})
-    output$eth_nma <- renderText("No forecast for the 2021 Belg season available from the NMA.")
-    output$copernicus <- renderText({c('<img src="','https://stream.ecmwf.int/data/gorax-blue-005/data/scratch/20201222-0920/c3/convert_image-gorax-blue-005-6fe5cac1a363ec1525f54343b6cc9fd8-b9q_5d.png"','width = "500px" height = "500px"', '>')})    
+    output$nmme <- renderText({c('<img src="','https://www.cpc.ncep.noaa.gov/products/international/nmme/probabilistic_seasonal/africa_nmme_prec_3catprb_FebIC_Mar2021-May2021.png"','width = "600px" height = "500px"', '>')})
+    #output$eth_nma <- renderText("No forecast for the 2021 Belg season available from the NMA.")
```
joseepoirier, line 86: I'd leave the note about eth_nma to show we've consulted them

dashboard/ui.R
```diff
@@ -50,10 +50,12 @@ fluidPage(
                                  tags$br(), 
                                  htmlOutput("iri"),
                                  htmlOutput("icpac"),
-                                 #img(src="icpac_rainfall_20201222_JFM2021.png", align = "center", width=600, height=600),                                 
-                                 htmlOutput("chc"),
+                                 #img(src="icpac_rainfall_20201222_JFM2021.png", align = "center", width=600, height=600),
+                                 img(src="eth_nma_precipitation_202101_Belg2021.png", align = "center", width=600, height=400),
                                  htmlOutput("nmme"),
-                                 textOutput("eth_nma")),
```
joseepoirier, line 56: i'd leave this in

dashboard/ui.R
```diff
@@ -50,10 +50,12 @@ fluidPage(
                                  tags$br(), 
                                  htmlOutput("iri"),
                                  htmlOutput("icpac"),
-                                 #img(src="icpac_rainfall_20201222_JFM2021.png", align = "center", width=600, height=600),                                 
-                                 htmlOutput("chc"),
+                                 #img(src="icpac_rainfall_20201222_JFM2021.png", align = "center", width=600, height=600),
+                                 img(src="eth_nma_precipitation_202101_Belg2021.png", align = "center", width=600, height=400),
```
joseepoirier, line 54: this does not display correctly. You have to save the PNG in the www folder.

joseepoirier (commented):
you're missing a file for the nma projections. the radio buttons are disabled. the trigger regions only show 2 as having met the trigger, which is not in agreement with what we summarised in the document. Let me know if you need help fixing this.

dashboard/ui.R
Tinkaa: Oops I forgot to add the png to Git🤦‍♀️

Tinkaa:
> the trigger regions only show 2 as having met the trigger, which is not in agreement with what we summarised in the document. 
Is this for FewsNet and the Feb-May period? Cause for me it is showing 4 regions for that period, which is in agreement with the document. Could it be that you looked at the Jan period, which is in the dashboard but not in the document? If not, can you send a screenshot so that I can check where the discrepancy comes from

dashboard/generate_subnational_map.R
Tinkaa: I haven't seen cases where we lack ML1s but do have ML2s, but I can imagine they will appear in the future thus prepared the code for that

Tinkaa: And so the row is not dropped if only one of the two is missing, separate dataframes are made for ML1 and ML2 which result in separate maps and trigger lists.

dashboard/generate_subnational_map.R
Tinkaa: No, the csv only includes Ethiopia at the moment. However, the script to compute the numbers is easily transferrable to other countries, but would need few changes on the dashboard side to built-in that flexibility. 
Added few details about required data file format

dashboard/server.R
```diff
@@ -12,10 +12,10 @@ source("generate_subnational_map.R")
...
-    fn_ml2 <- ipc_indices_data %>% filter(Source == 'FewsNet') %>% select(ML2_period) %>% unique() %>% as.character()
-    gbl_ml1 <- ipc_indices_data %>% filter(Source == 'GlobalIPC') %>% select(ML1_period) %>% unique() %>% as.character()
-    gbl_ml2 <- ipc_indices_data %>% filter(Source == 'GlobalIPC') %>% select(ML2_period) %>% unique() %>% as.character()
+    fn_ml1 <- ipc_indices_data_latest %>% filter(source == 'FewsNet') %>% select(period_ML1) %>% unique() %>% as.character()
+    fn_ml2 <- ipc_indices_data_latest %>% filter(source == 'FewsNet') %>% select(period_ML2) %>% unique() %>% as.character()
+    gbl_ml1 <- ipc_indices_data_latest %>% filter(source == 'GlobalIPC') %>% select(period_ML1) %>% unique() %>% as.character()
+    gbl_ml2 <- ipc_indices_data_latest %>% filter(source == 'GlobalIPC') %>% select(period_ML2) %>% unique() %>% as.character()
 
     # conditionally select correct map
     output$trigger_map <- renderPlot({
```
joseepoirier, line 21: @Tinkaa Here's why your radio buttons don't work. You removed the "country" variable from the csv.

joseepoirier (commented):
need to add the "country" variable back into the csv. also remove the unnamed columns with rownumbers

joseepoirier:
@Tinkaa Merging per our conversation re: restructuring the repo and to allow me to make changes to the UI requested by management

PR #39, opened 2021-02-22
Directory restructure

turnerm:


turnerm:
Merging now so that people can keep working. Things left to do:
- `FewsNetWorldPop` processed data should go in the `processed` directory
- all scripts in `analysis` need to be updated to point to the new data directory (except for the R scripts in Malawi)

PR #41, opened 2021-02-24
Mwi precipobs

Tinkaa:
For MWI analysis I created a notebook containing the computation of historical observations of below-average rainfall, for monthly as well as seasonal (3-monthly) periods. 
And did a quick contingency table of those observations with the preliminary list of dry spells. 
Would one of you have time to have a good look at it? Especially the historical seasonal below-average rainfall observations I am scared for methodological mistakes


hannahker:
@Tinkaa I looked through the ```mwi_obsprecip``` notebook and have some general questions and thoughts. I'm not super experienced with processing gridded data in Python, but conceptually the workflow makes sense to me. Out of curiosity, why is there the need to compare against both monthly and seasonal observed precipitation? 

Regarding terciles: So for the CAMP_OPI for example, terciles are based on the overall distribution of global precipitation measurements from 1981-2010? If so, I thiink this would be why your calculated fractions aren't summing to 0.33, as you're clipping to just Malawi and there's no guarantee that the distribution of values for Malawi will match the global distribution of values?

And to some of your questions: 

- Is it valid to say there was below average rainfall in an admin2 if any cell touching that region had below average rainfall?Ideally I think you would want to classify based on the cells that cover the majority of the admin region's area. I would say that you definitely don't want the classification to be made potentially based on just one cell when you're looking at something like admin 2 regions with 0.5 deg resolution. 

- Is it essential to include a dry mask?
I don't really understand the point of the dry mask. Isn't the aim to identify dryness?

Tinkaa:
@hannahker Super thanks!
> why is there the need to compare against both monthly and seasonal observed precipitation?
Because we want to look into using monthly as well as seasonal forecasts and thus want to know if there is a correlation with the rainfall in those periods and dry spells
> terciles are based on the overall distribution of global precipitation measurements from 1981-2010? If so, I thiink this would be why your calculated fractions aren't summing to 0.33, as you're clipping to just Malawi and there's no guarantee that the distribution of values for Malawi will match the global distribution of values?
As far as I understand my own code, I am computing the terciles per cell, i.e. each cell has its own tercile boundary value. So in that case it should be 0.33? Correct me if I am wrong. I think the wrong fraction was because I was computing it on the processed dataframe instead of the raster data. Changed that now and seems correct now
>Ideally I think you would want to classify based on the cells that cover the majority of the admin region's area.
Yes good point! I am gonna change that for the CHIRPS data now. For CAMS-OPI the resolution is too low for that, so leaving that for now, especially cause I am thinking we should maybe also use CHIRPS for the monthly values. 
>I don't really understand the point of the dry mask. Isn't the aim to identify dryness?
Yes, or well areas where it is generally speaking dry. What I understand is that in those cases there is dryness anyway so the differences can be small and might classify it as a deviation from the normal while actually those changes don't influence the e.g. cropping pattern. Not 100% sure though. But maybe for short-term analysis will leave it out.

hannahker:
@Tinkaa my comments on ```mwi_obsprecip_corr_dryspells.ipynb```. Thanks for adding in the context at the top of the notebook and clear questions 😄 

Is the code to aggregate dry spells to seasons correct? Do you agree with the methodology to aggregate dry spells to seasons or would do it differently?

- The only potential problem that I can think of is the fact that some dry spells cover multiple months (29 are longer than 30 days). Although I think this isn't hugely impactful as only three of these are longer than 45 days (if we were to use the criteria that a dry spell should take up the majority of a month).

Is there an analysis I am missing?

- Nothing that I can think of. So to be clear this analysis is answering the question: Is the presence of below avg seasonal rainfall a predictor of whether or not a dry spell will occur in a given adm2 region for that season? And: For each adm2 is the number of dry spells correlated with the number of seasons with below avg rainfall?

Would you say that based on the current analysis it is valid to conclude that there is no signal in below average seasonal rainfall to predict dry spells?

- Yeah, looks like it 😢

Tinkaa:
@hannahker finally managed to get the chirps-gefs code ready for review, so it would be great if you could have a look! 
The `malawi/notebooks/mwi_chirps_gefs.ipynb` notebook downloads the data and computes ADM2 statistics. The `malawi/notebooks/mwi_chirpsgefs_corr_dryspells.ipynb` notebook analyzes the correlation between the forecast data and observed dry spells. This is a really long notebook with loads of graphs, so sorry for that. However, review till cell 70 should be more than enough, and you def don't have to look through it very detailed. 
I tried to add questions, but most important check would be to see if I am not making any stupid mistake with processing and aggregating data.

hannahker:
@Tinkaa some comments for the ```malawi/notebooks/mwi_chirps_gefs.ipynb``` 😃 
- So it makes sense that ```df_rain_filled``` has less rows (688) than ```df_rain``` (704) right, because you're dropping cases where both the onset and cessation dates are null?
- What's the justification for manually assigning the onset and cessation dates with Nov 1/July 1 where they are null? Are these conservative estimates for where there isn't the data available?  
- Some of the long method chaining is a biiiit hard to follow (eg. cell 9, 22), but just a style thing so not critical :) 
- For the ```+timedelta(days=14)```,  now I'm overthinking what a 15-day forecast really means... my initial reaction is that you would add 15 days rather than 14, as the first day would not be the current day. Is there a resource that we could verify this with?

hannahker:
@Tinkaa and on ```mwi_chirpsgefs_corr_dryspells.ipynb```: 
- My first reaction is that this is a really useful reference as I'm getting in to evaluating/comparing GloFAS forecasting :) 
- For ```df_chirpsgefs```, I am a bit confused trying to figure out how many rows there should be. Shouldn't there be one entry per adm2 for every dry spell day? So 3358 x 32? Which means that the 77184 rows is a bit low. But I could easily have missed something here.
- The joinplots are really interesting and useful! Although I'm not entirely clear on what the difference is between each of them. So the first one is essentially saying that CHIRPS GEFS underestimates precipitation more when the observed precipitation is higher?
- I'm not sure that the next plot on obs and forecasted rainfall in Balaka says very much. Maybe this could be subset by years?
- I'll admit that I didn't fully follow everything at the end of the notebook, but as I understand we're comparing between forecasted and observed dry spells at the level of a day - ie. how well does a dry spell day (rather than a dry spell itself as an entity) correlate between the two datasets? In this case it perhaps makes sense that the correlations are poor because exact day-by-day matching is a pretty strict requirement for dry spell correlation. Would it also be interesting to look at cases where dry spells are something like at least 50% overlapping?

hannahker (approved):
Looks good to me, with the comments that I just added 😄

Tinkaa:
@hannahker 
> some comments for the malawi/notebooks/mwi_chirps_gefs.ipynb 😃
A belated but very much meant big thanks!

> So it makes sense that df_rain_filled has less rows (688) than df_rain (704) right, because you're dropping cases where both the onset and cessation dates are null?
Yes exactly, this happens during 2020 for a subset of adm2s. Comment added

> What's the justification for manually assigning the onset and cessation dates with Nov 1/July 1 where they are null? Are these conservative estimates for where there isn't the data available?
Yes exactly. Done only for making sure to download all the data, in the end only dates which are within the season are selected. Comment added

> Some of the long method chaining is a biiiit hard to follow (eg. cell 9, 22), but just a style thing so not critical :)
Agreed. Leaving it for now, but trying to pay attention to it for future

>For the +timedelta(days=14), now I'm overthinking what a 15-day forecast really means... my initial reaction is that you would add 15 days rather than 14, as the first day would not be the current day. Is there a resource that we could verify this with?
What do you mean with "as the first day would not be the current day"? Cause I think the first day is the first day of the forecast, so then it would end 14 days after? But I am touch with one of the people involved with CHIRPS-GEFS now, so will ask her to be sure :)

Tinkaa:
@hannahker 
> My first reaction is that this is a really useful reference as I'm getting in to evaluating/comparing GloFAS forecasting :)
 
🤩


> For df_chirpsgefs, I am a bit confused trying to figure out how many rows there should be. Shouldn't there be one entry per adm2 for every dry spell day? So 3358 x 32? Which means that the 77184 rows is a bit low. But I could easily have missed something here.

Such sharpness! Got me thinking again, butt I am pretty sure it is because most of the time ADM2's end their rainy season earlier (or start later), and those are filtered out. So the total number of unique dates is 3358, but not all adm2's contain these dates

> The joinplots are really interesting and useful! Although I'm not entirely clear on what the difference is between each of them. So the first one is essentially saying that CHIRPS GEFS underestimates precipitation more when the observed precipitation is higher?

They are not different, only zoomed in, since we are focussing on small amounts of precipitation. Yes indeed saying underestimates when higher, and at the same time overestimating when very little precipitation

> I'm not sure that the next plot on obs and forecasted rainfall in Balaka says very much. Maybe this could be subset by years?

Agree, removing

> I'll admit that I didn't fully follow everything at the end of the notebook, but as I understand we're comparing between forecasted and observed dry spells at the level of a day - ie. how well does a dry spell day (rather than a dry spell itself as an entity) correlate between the two datasets? In this case it perhaps makes sense that the correlations are poor because exact day-by-day matching is a pretty strict requirement for dry spell correlation. Would it also be interesting to look at cases where dry spells are something like at least 50% overlapping?
We are looking at whether at least one day of a dry spell in one of the two data sets overlaps with a dry spell in the other dataset. So this is the most loose definition we can have. Completely at the end, we look indeed at very strict definition of the dry spell starting on the date of the forecast, but concluded this is probably too strict.

Tinkaa:
As I think most important files are reviewed, I will merge it into develop.

PR #42, opened 2021-02-25
Mwi hist dry spells

joseepoirier:
@hannahker @Tinkaa 

I used the 5deg resolution data by loading the `data` and `data_max_values` RDS files.

This code identifies the onset and end of each rainy season (runs No-May ish in MWI), then flags every dry spell within a rainy season per adm2. 

Please double check if there is a reason no region has dry spells longer than 15 days: that shouldn't be the case. I'll take a closer look tomorrow morning as well. Thanks!


Tinkaa:
Just a practical note, I think you want to merge this into develop instead of main cause that follows the new repo structure (though I see you also merged main into it)

PR #43, opened 2021-02-26
GloFAS download and processing for Bangladesh

turnerm:
- created script `src/indicators/flooding/glofas.py` that downloads and processes GloFAS reanalysis, reforecast, and forecast
  - note that for the forecast, it downloads the data by year, so it's meant more for getting the data in bulk rather than pinging the forecast every day when we will need it for a trigger -- would be best to write another routine for that (in the same file though)
- I moved all the previous GloFAS script back to the `analysis` folder as they are more one-off analyses rather than something for the pipeline
  - Ignore the changes to `analyses/bangladesh/trigger_development/GetGLOFAS_Data.py`; I am just reverting it back to its original form
- At the moment, the download / processing script takes both the country name and iso3 as arguments. This is because, to make the filenames descriptive, I wanted to include the iso3 in the filenames. But for the directory structure we are currently using the full country names. I think we should probably switch to using iso3 for everything instead (because once we get to countries with spaces in the name things will get messy fast...) and thus using iso3 for the filenames was an attempt at forward compatibility
- The main script to run is `src/bangladesh/get_glofas.py`. At the end of the download / processing steps I've added a few lines to show how to begin with an analysis, these should probably be moved elsewhere.


hannahker:
This looks really great! Thanks for adding in that example analysis bit - really helps to understand how to put this into use. This definitely seems like something that would be great to publish as a resource for others to use too :) 

Do you think in the future it would be good to read in the stations (FFWC or otherwise) from an external config file?

hannahker (approved):

src/bangladesh/get_glofas_data.py
```diff
@@ -0,0 +1,94 @@
...
+    year = "2020"
+    # TODO: add interpolation to make sure there aren't any missed dates
+    da_reanalysis = glofas_reanalysis.read_processed_dataset(
+        country_name=COUNTRY_NAME, country_iso3=COUNTRY_ISO3
+    )[station].sel(time=slice(year, year))
+    da_forecast_720 = (
+        glofas_forecast.read_processed_dataset(
+            country_name=COUNTRY_NAME, country_iso3=COUNTRY_ISO3, leadtime_hour=720
+        )[station]
+        .shift(time=30)
```
hannahker, line 75: What is this shift doing?

src/bangladesh/get_glofas_data.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if process:
+        glofas_reanalysis.process(country_name=COUNTRY_NAME, country_iso3=COUNTRY_ISO3)
+        glofas_forecast.process(country_name=COUNTRY_NAME, country_iso3=COUNTRY_ISO3)
+        glofas_reforecast.process(country_name=COUNTRY_NAME, country_iso3=COUNTRY_ISO3)
+
+    # Start of an analysis
+    # Read in the data for a station, and select the year 2000
+    station = "Bahadurabad"
+    year = "2020"
```
hannahker, line 66: Maybe need to play around with the code some more, but I'm a bit unclear on how you would download/process the data by year. So the methods in ```glofas.py``` will query for all data between ```year_min``` and ```year_max```, and then from there you would select a single year of interest as you've done below?

src/bangladesh/get_glofas_data.py
turnerm: It shifts the time coordinate by 30, since this is the dataset with the 30 day leadtime, and there is a forecast for each day. The reforecast data will need to be interpolated before such a shift can be done since it's only run every couple of days.

src/bangladesh/get_glofas_data.py
turnerm: I just chose a single year for the "analysis" here because otherwise the plot would get too crowded.

For the download, I did have an option to specify a year range, but removed it because I didn't think of a good use case. The download option uses a cache, i.e. if the data exists it doesn't download it again, so you wouldn't have to specify a year for that reason. I can definitely add it back in if you think it's useful though. 

For the processing I could add it in as well, but then I think it should also be added to the output filename just so it's clear which parameters were used to create the data.

turnerm:
Merged without adding a year range argument to download / process methods but happy to do that in a new branch

PR #46, opened 2021-02-27
Mwi hist dry spells

joseepoirier:
cleaned up a bit the code to compute dry spells. still surprised that there would not be dry spells longer than 15 days, but starting to wonder if it's an artefact of the definition: as soon as there is a single rainy day with >= 2mm in the middle of the dry spell, the dry spell ends because the rolling sum has breached the threshold of 2mm and is no longer considered a dry spell. 

Still, please poke around to see if we're artificially creating an upper limit to the duration of dry spells. 
Tinka: will add total rainfall back, realised overnight it summed over the confirmation dates and not the whole dry spell.


Tinkaa:
Hmm, the max 15 days sounds odd to me but might very well be. Is this only in the rainy season or whole year? Cause if whole year I would say >15 day dry spells are even more likely

joseepoirier:
This code only looks at dry spells during the rainy season.

On 2/28/21, 9:56 AM, "Tinka" <notifications@github.com> wrote:


Hmm, the max 15 days sounds odd to me but might very well be. Is this only in the rainy season or whole year? Cause if whole year I would say >15 day dry spells are even more likely

analyses/malawi/scripts/mwi_chirps_dry_spell_detection.R
```diff
@@ -85,144 +99,267 @@ for (i in seq_along(1:nbr_layers)) {
...
-                                              median_duration = mean(duration_days),
-                                              min_duration = min(duration_days),
-                                              max_duration = max(duration_days))
+dry_spells_during_rainy_season_list <- dry_spells_during_rainy_season_list %>% 
+                                          left_join(mwi_adm2_ids, by = c('pcode'= 'ADM2_PCODE')) %>%
+                                          dplyr::select(pcode, ADM2_EN, dry_spell_first_date, dry_spell_last_date, dry_spell_duration, dry_spell_rainfall)
+
+write.csv(dry_spells_during_rainy_season_list, file = paste0(data_dir, "/processed/malawi/dry_spells/dry_spells_during_rainy_season_list.csv"), row.names = FALSE)
+
+# summary stats per region
```
hannahker, line 309: As I understand here region is referring to admin1? I would just be somewhat careful of the accuracy of results aggregated to a higher admin level than the original results are based on. As the results you have are sensitive to the shape of the admin2 boundaries, it's not a given that the results would be the same as if you had processed the chirps data and calculated the dry spells straight from the admin1 boundaries. 

Looks like just an exploration here, but maybe something to keep in mind.

hannahker (commented):
Looks good to me @joseepoirier 💃 I tried to go fully end-to-end in rerunning the code but got a bit impatient with the ```data_max_values``` calculation... I'm still convinced that GEE has a great solution for us there! But seems like it's not anything urgent right now.

analyses/malawi/scripts/mwi_chirps_dry_spell_detection.R
joseepoirier: Good point. Yes I was computing at region (admin1) level simply to verify a WFP analysis that saw most dry spells in the South. We should keep it in mind if we want to do anything at adm1, thanks for raising.

joseepoirier:
> Looks good to me @joseepoirier 💃 I tried to go fully end-to-end in rerunning the code but got a bit impatient with the `data_max_values` calculation... I'm still convinced that GEE has a great solution for us there! But seems like it's not anything urgent right now.

Ha, yes. Sorry - I had saved the RDS for `data` and `data_max_values` objects for that reason: it takes time to create them. Last night I added years 2000-2009 so it takes much longer, too. When reviewing the code I would recommend loading the RDS files instead of rebuilding the two R objects.

hannahker:
Yes, when reading in the RDS file ```data_max_values_2000_2020_r5.RDS``` I'm only getting data for the first 4 days of 2000?

joseepoirier:
The analysis is complete and the write-up is ready for consumption & feedback!

hannahker (approved):
@joseepoirier I'm in awe of how quickly you put that write-up together! I think the write up is quite clear and it's good to have all the definitions (dry spell, onset, cessation, etc.) so clearly described. I also quite like the validation against the WFP report. I appreciate that you're trying to send this out today, so just confirming that I think it's good to go 👍 

Some suggestions for 'nice-to-haves':
- Some visualisation might be helpful for people to get a clearer sense of what's going on with the data. It could also be a good validation of your analysis to map out some of the variables (eg. avg dry season start date, nbr of dry spells), as ideally you would see that things are to some extent spatially autocorrelated. Or it could also be a good exercise in finding some outliers to investigate further. 
- A general sense of the trend over time? It might be interesting to know if the occurrence of dry spells is becoming more frequent over time. 

As a side note, I really like that format of communicating results. Would be interested to talk more about cases where we could use this kind of format vs slides.

joseepoirier:
> @joseepoirier I'm in awe of how quickly you put that write-up together! I think the write up is quite clear and it's good to have all the definitions (dry spell, onset, cessation, etc.) so clearly described. I also quite like the validation against the WFP report. I appreciate that you're trying to send this out today, so just confirming that I think it's good to go 👍
> 
> Some suggestions for 'nice-to-haves':
> 
> * Some visualisation might be helpful for people to get a clearer sense of what's going on with the data. It could also be a good validation of your analysis to map out some of the variables (eg. avg dry season start date, nbr of dry spells), as ideally you would see that things are to some extent spatially autocorrelated. Or it could also be a good exercise in finding some outliers to investigate further.
> * A general sense of the trend over time? It might be interesting to know if the occurrence of dry spells is becoming more frequent over time.
> 
> As a side note, I really like that format of communicating results. Would be interested to talk more about cases where we could use this kind of format vs slides.

Thank you so much Hannah. I agree that visualisations would be very effective - I just ran out of time but will try to sneak in a few because you are right, the story is incomplete without them. 

Happy to discuss when our team thinks using Markdown/HTML vs slides makes most sense.

joseepoirier:
Per our conversation I am merging this branch into dev.

PR #51, opened 2021-04-13
Bangladesh trigger part 1

turnerm:
- Added Bahadurabad GloFAS station, re-ran processing to include it
- Added a notebook for Bangladesh flooding pre-activation trigger analysis. Used to generate figures for [this](https://docs.google.com/presentation/d/1H1ofAybnseKoiSHvAmsE24S4W7hZhrb8dmd-ebqTyq4/edit?usp=sharing) slide deck.


analyses/bangladesh/trigger_development/bgd-trigger.md
```diff
@@ -0,0 +1,569 @@
...
+reload(rd)
+
+mpl.rcParams['figure.dpi'] = 300
+```
+
+### Create GloFAS objects
+
+```python
+da_glofas_reanalysis = rd.get_glofas_reanalysis()
+da_glofas_forecast = rd.get_glofas_forecast()
```
hannahker, line 22: Hmm I'm getting an error on this line...

```
AttributeError                            Traceback (most recent call last)
<ipython-input-2-b39b49459c1d> in <module>
      1 da_glofas_reanalysis = rd.get_glofas_reanalysis()
----> 2 da_glofas_forecast = rd.get_glofas_forecast()
      3 da_glofas_forecast_summary = rd.get_da_glofas_summary(da_glofas_forecast)
      4 da_glofas_reforecast = rd.get_glofas_reforecast()
      5 da_glofas_reforecast_summary = rd.get_da_glofas_summary(da_glofas_reforecast)

~\Desktop\pa-anticipatory-action\analyses\bangladesh\trigger_development\read_in_data.py in get_glofas_forecast()
     43     }
     44     da_glofas_forecast_dict = shift_dates(da_glofas_forecast_dict)
---> 45     return convert_dict_to_da(da_glofas_forecast_dict)
     46 
     47 

~\Desktop\pa-anticipatory-action\analyses\bangladesh\trigger_development\read_in_data.py in convert_dict_to_da(da_glofas_dict)
     99         dims=["leadtime_hour", "number", "time"],
    100         coords=dict(
--> 101             number=da_glofas_dict[120].number,
    102             time=time,
    103             leadtime_hour=list(da_glofas_dict.keys()),

~\anaconda3\envs\bang_floods\lib\site-packages\xarray\core\common.py in __getattr__(self, name)
    226                 with suppress(KeyError):
    227                     return source[name]
--> 228         raise AttributeError(
    229             "{!r} object has no attribute {!r}".format(type(self).__name__, name)
    230         )

AttributeError: 'DataArray' object has no attribute 'number'
```

analyses/bangladesh/trigger_development/read_in_data.py
```diff
@@ -0,0 +1,173 @@
...
+from scipy.stats import norm, pearsonr
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+path_mod = f"{Path(os.path.dirname(os.path.realpath(''))).parents[1]}/"
+sys.path.append(path_mod)
+
+from src.indicators.flooding import glofas
+from src.bangladesh import get_glofas_data as ggd
```
hannahker, line 15: Out of curiosity, what do you think the value is in storing the global variables for BGD in ```get_glofas_data``` vs this script itself?

analyses/bangladesh/trigger_development/read_in_data.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+        for leadtime_hour, da in da_dict.items()
+    }
+
+
+def interp_dates(da_dict):
+    return {
+        leadtime_hour: da.interp(
+            time=pd.date_range(da.time.min().values, da.time.max().values),
+            method="linear",
```
hannahker, line 76: On looking back, my use of linear interpolation previously wasn't a very deliberate decision. Do you think it's worth considering another method, like cubic?

hannahker (commented):
I scanned through the ```.md``` and all looks good but couldn't rerun because of that error. Do you have any idea what's going on there?

analyses/bangladesh/trigger_development/read_in_data.py
```diff
@@ -0,0 +1,173 @@
...
+        leadtime_hour: da_glofas.reindex({"time": time})
+        for leadtime_hour, da_glofas in da_glofas_dict.items()
+    }
+
+    data = np.array([da_glofas.values for da_glofas in da_glofas_dict.values()])
+    return xr.DataArray(
+        data=data,
+        dims=["leadtime_hour", "number", "time"],
+        coords=dict(
+            number=da_glofas_dict[120].number,
```
hannahker, line 100: I'm a bit unclear on what ```number``` is here...

analyses/bangladesh/trigger_development/bgd-trigger.md
turnerm: Hm that is strange. Are you using the latest data? It's referring to the following dataset:
`processed/bangladesh/GLOFAS_Data/bgd_cems-glofas-forecast_0120.nc`
If you open the data array directly, do you see that it has the variable `number`?

analyses/bangladesh/trigger_development/read_in_data.py
turnerm: Good question, so at the moment `get_glofas_data` is kind of acting as the main entry script, and thus seems like the most natural place to store this information, which probably shouldn't be unnecessarily duplicated. But anyway with the Nepal flooding PR I refactored the GloFAS class so I don't think these variables will be required anymore.

analyses/bangladesh/trigger_development/read_in_data.py
turnerm: No I think linear was a great choice! I don't think there's any compelling argument for using a more complex interpolation scheme, so I think it's safest to go with the simplest option. Also with other interpolation methods you can sometimes end up with wacky artifacts.

analyses/bangladesh/trigger_development/read_in_data.py
turnerm: Oops good point, it's the ensemble member number, I will clarify this

analyses/bangladesh/trigger_development/bgd-trigger.md
hannahker: Arg ok yes so when I'm opening ```processed/bangladesh/GLOFAS_Data/bgd_cems-glofas-forecast_v3_lt05d.nc``` I do see ```number```... I guess something about my xarray setup really doesn't like making things easy for me 😢

analyses/bangladesh/trigger_development/bgd-trigger.md
```diff
@@ -0,0 +1,575 @@
...
+
+```
+
+### Full 30 year figure
+
+```python
+fig, ax = plt.subplots(figsize=(20,10))
+ax.plot(df_final.index, df_final['observed'], 'o', 
+        label='FFWC observations')
+ax.plot(df_final.index, df_final['ffwc_5day'], 's',
```
hannahker, line 479: Also getting this... is there something missing from my input ffwc data?

```
KeyError                                  Traceback (most recent call last)
<ipython-input-23-f3e0cbebef3d> in <module>
      2 ax.plot(df_final.index, df_final['observed'], 'o', 
      3         label='FFWC observations')
----> 4 ax.plot(df_final.index, df_final['ffwc_5day'], 's', 
      5         alpha=0.5, label='FFWC 5 day forecast')
      6 idx_event = df_final['event'] == True

~\anaconda3\envs\bang_floods\lib\site-packages\pandas\core\frame.py in __getitem__(self, key)
   2900             if self.columns.nlevels > 1:
   2901                 return self._getitem_multilevel(key)
-> 2902             indexer = self.columns.get_loc(key)
   2903             if is_integer(indexer):
   2904                 indexer = [indexer]

~\anaconda3\envs\bang_floods\lib\site-packages\pandas\core\indexes\base.py in get_loc(self, key, method, tolerance)
   2891                 return self._engine.get_loc(casted_key)
   2892             except KeyError as err:
-> 2893                 raise KeyError(key) from err
   2894 
   2895         if tolerance is not None:

KeyError: 'ffwc_5day'
```

analyses/bangladesh/trigger_development/bgd-trigger.md
turnerm: Yeah this problem really sounds awful! I will message you with some more ideas on how to troubleshoot

analyses/bangladesh/trigger_development/bgd-trigger.md
turnerm: Oops good catch! This error came from when I updated to using days instead of hours for leadtimes. I had fixed it in an upstream branch but not here. Should be fixed with the next commit!

hannahker:
@turnerm I think this should probably be good to merge at this point! I think lots of this will be transferable to MWI so I might follow up with some more specific questions soon as I dig into this more closely.

PR #52, opened 2021-04-15
Nepal GloFAS

turnerm:
Thought I would get a head start and download the GloFAS data for Nepal. To do this:
- Added a `nepal` directory with a script to download and process the GloFAS data
- Added method to get download area from a polygon (using the bounds box). Since for Nepal we want stations in both the east and west, and it's not such a huge country, I thought it might be easiest to just get the whole country in one go. 

In the process, also made some changes to the GloFAS class: 
- Created an `Area` class that deals with the creation and formatting of the shape passed to the GloFAS query
- Improved type checking (check for types within dicts & lists)
- Make GloFAS class intantiation fully generic: no longer require leadtime hours or station list to create a glofas object instance. Only pass those parameters when they're needed for download and processing.


src/indicators/flooding/glofas/area.py
```diff
@@ -0,0 +1,47 @@
...
+        self.south = south
+        self.east = east
+        self.west = west
+
+    def list_for_api(self):
+        return [self.north, self.west, self.south, self.east]
+
+
+class AreaFromStations(Area):
+    def __init__(self, stations: Dict[str, Station], buffer: float = 0.5):
```
hannahker, line 21: I'm wondering if a default buffer of 0.5 degrees might be a bit larger than one would ever need?

hannahker (approved):
This all looks good as far as I can tell. This is forcing me to dig out my memory of OOP from my Intro to Programming days... Just added the comment about default buffer size. 

My only other thought was that it seems like there's decent repetition between the ```get_glofas_data.py``` files for both Bangladesh and Nepal. Would it make sense to generalize this further?

src/indicators/flooding/glofas/area.py
turnerm: That's a good point, I'll change it to 0.1

turnerm:
> My only other thought was that it seems like there's decent repetition between the get_glofas_data.py files for both Bangladesh and Nepal. Would it make sense to generalize this further?
That's a good idea. I'll be converting these two scripts into top-level entrypoint scripts for the AA repo refactor soon, so I think I will defer this refactor to then, if that's okay.

src/indicators/flooding/glofas/area.py
hannahker: Oh yeah, fair! I didn't think about it in terms of GloFAS resolution.

PR #53, opened 2021-04-19
Bangladesh flood skill

turnerm:
Continuing Hannah's skill notebook:
- updated to use more general methods to read in the GloFAS data
- added full year calculation (to compare with rainy season)
- added rank histogram to examine bias
- converted notebook to markdown

The rank histograms don't look great, I'm worried that I made a mistake somewhere.

(Also: there is something going on with the 30-day reforecast processed data. Investigating in parallel)


hannahker (approved):
Finally got things working from my end with Insync :) All looks great to me so far. I'll probably have some more specific questions as we try to reuse some of this for MWI, but I think definitely good to merge into ```develop```.

I'm having a hard time interpreting the rank histograms... will follow up with you over Slack on that next week. Also it seems interesting/surprising on the last plots of percent mean error that the direction of error during the rainy season shifts between positive and negative across the various lead times (for v3).

PR #54, opened 2021-04-19
Eth blog post

joseepoirier:
@Tinkaa Can you double check the data are correct? Some regions at virtually no food insecurity (IPc3+) over the 12-year period while a few had 100% of their population in IPC3+ at times. I want to make sure the data I read in is accurate and that I don't introduce an error when computing the percentages.


Tinkaa:
It is slightly hard to be 100% sure it is correct, but I would say so. 
The historical figures correspond with the trends in this notebook (even though that uses the worldpop instead of admin methodology)
Moreover, checking the png files on the FewsNet website for 2015 and 2017, the spatial patterns correspond with the percentage patterns

I do think though we should be careful having the computation of data in different places. For now for the blog post is fine, but I should incorporate the percentage computation in the `process_fewsnet_subnatpop.py` if this shows we will continue using that methodology.

joseepoirier:
Ok my spot checks also looked good but feel more confident having had another set of eyes. Agreed about having the computations in the origin code.

Tinkaa:
@joseepoirier since the blogpost is finished, should we merge this? 
I am doubting a bit if we want to include the plots on the branch, but at the same time I agree with you it is nice that people can browse

Edits I made
- Include FI map for Global IPC
- Increase font sizes

Tinkaa:
Yes I think it is better to put them on the GDrive as not to clutter `develop`
I made a plots directory in `public/processed/eth` and changed the paths in the R file to point to those. 
Will merge now.

PR #56, opened 2021-04-22
GloFAS model update

turnerm:
The new GloFAS model, while not yet operational, is available for download! Changes are:
- update the GloFAS class to have a model version option
- change leadtimes to be in days instead of hours (except for the query itself of course)
I've also updated all of the relevant file structures.
- create a separate v3 and v2 download script for Bangladesh (to preserve how all the v2 data was downloaded)
- made splitting by month optional (for backward compatibility -- GloFAS increased the query size! :partying_face: )

These changes break things in some open branches, but since they are urgent I've implemented them now, I will merge this branch into the others and propagate the changes through.


hannahker (approved):
Looks great to me, and I like the adjustment to specify leadtime in days rather than hours. The only thing that I found slightly confusing was the naming of ```get_glofas_data_v2.py``` (since without knowing one might assume that this is v2 of the script, rather than the GloFAS model). But not a big deal.

src/indicators/flooding/glofas/glofas.py
```diff
@@ -31,7 +33,7 @@ def __init__(
         cds_name: str,
         dataset: List[str],
         dataset_variable_name: str,
-        system_version_minor: int = None,
+        system_version_minor: Dict[int, int],
```
hannahker, line 36: Should have picked this up in a past review, but what is ```system_version_minor```?

hannahker:
@turnerm confirming that I looked through the last commit as well 👍

turnerm:
> The only thing that I found slightly confusing was the naming of get_glofas_data_v2.py (since without knowing one might assume that this is v2 of the script, rather than the GloFAS model).

That's a really good point! I'll rename the file to get_glofas_data_old_model.py.

src/indicators/flooding/glofas/glofas.py
turnerm: Good question! So when you make the query to CDS you have to specify the model version, which has the format {major}.{minor}. The major version is 3 for the new model, 2 for the old one. Minor is almost always 1, except for the old mode reforecast, where it is 2  -- hence the need to turn it into a dictionary, where the key is the major version number. 

But anyway you're right that this point needs more clarification, so I've added a docstring. In general this class could use a few more docstrings...!

PR #57, opened 2021-04-23
Feature/gee

Tinkaa:
@hannahker  I managed to use the `gee_zonal_stats_ts.py` which is great :DD I tried to generalize it a bit, though I think it would be better if we take all the stuff now in main to a country-specific script. And then I would suggest to move the gee script to `utils_general` instead of `indicators/drought`. What do you think?


hannahker (approved):
@Tinkaa this looks great to me. Thanks for making those adjustments to generalize things further. I've also moved it to ```utils_general```, which I agree makes more sense. I've deleted the old chirps one as well since I think it's redundant. I'd like to also add in a function to clean up the ```.csv``` results and combine them to a single file (with the other R script is doing, but not in a totally generalized way), so that there is one single, clean csv as output. But something for the future probably...

PR #58, opened 2021-04-23
make plot of chirps monthly precipitation

Tinkaa:
@hannahker I basically copied your monthly precipitation code from mwi to ssd, so thanks for making that so easy to use! The only struggle I encountered was that the value for `names <- df[(n-8),]` and `df <- df[1:(n-12),]`  seem to be different for mwi and ssd due to a different structure of the data (for ssd had to adjust it to `names <- df[(n-5),]` and `df <- df[1:(n-13),]` ). 
Do you have any idea what might causes this difference in file structure (probably caused by the gee script?) and how we could generalize it?

### DIFF
--- a/analyses/ssd/scripts/ssd_plot_monthly_precipitation.R
+++ b/analyses/ssd/scripts/ssd_plot_monthly_precipitation.R
@@ -6,24 +6,17 @@ library(tibble)
 library(lubridate)
 library(zoo)
 library(raster)
-# library(ggcorrplot)
 
 # -------------------------------------------------------------------------
-# Exploring agricultural stress and related impacts in Malawi
+# Exploring monthly rainfall in South Sudan
 # -------------------------------------------------------------------------
 
 
 # Setup -------------------------------------------------------------------
 
 data_dir <- Sys.getenv("AA_DATA_DIR")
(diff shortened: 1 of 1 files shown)


hannahker:
@Tinkaa yeah that was the bit that I hard coded based on the MWI outputs - glad you were able to adjust it for SSD and agree that it would be great to have more generalized. 

You probably ran into a problem because of differences in the ordering and number of columns in the admin boundary shapefiles between these two countries. The output files from the gee script have the admin file column names added in as the last set of columns. And so I was referencing the location of the boundary names to set as headers, and then dropping all of the other ones. 

To generalize maybe you'd have to specify the name (or position) of the boundary name column as an input, and then we could get it to drop all other columns that don't match the naming convention of the gee zonal stats output?

Tinkaa:
Aah yes that makes sens. Boundary name column you mean the column where the adm is in right? So for MWI adm1 it would be ADM1_EN? Cause that is already defined in the `config.yml` as `shp_adm1c` so we could use that? 
Would it make sense to do that directly in the python script?

hannahker:
@Tinkaa yes, exactly! So I think pulling that from the ```config.yml``` would work well then.

PR #59, opened 2021-04-26
Merge gh-pages to develop

Tinkaa:
Proposing to merge the ability to show the markdown notebooks with an url to develop. 
@hannahker I haven't looked into how this exactly works, so indicate if that doesn't make sense.


hannahker:
@Tinkaa I think this would be fine and I could just update the branch that is deployed via Github pages. I guess the only thing is that I've been merging to this branch much more liberally than to ```develop``` to enable access to an HTML Rmd output for an in-progress analysis. But I think not a big deal and I can just get better about getting things merged to ```develop``` more quickly!

hannahker (approved):

Tinkaa:
Yeah was thinking about that as well. But lets try if it works to directly merge to `develop` (would say so as that is already an extra branch before `main`) and else we can revive this branch again

PR #61, opened 2021-04-27
Malawi GloFAS

turnerm:
Script to download GloFAS data for Malawi. Merging into `bgd-skill` branch because I needed some fixes to the GloFAS library that were on that branch. (Will switch to develp once the skill branch is merged)


hannahker (approved):
Looks good to me. So this is for the most part boilerplate for each new country, right?

turnerm:
> So this is for the most part boilerplate for each new country, right?

Yes, although it should probably be switched to use a config file (I'll try to take care of this when I tackle the task of creating top-level scripts)

PR #63, opened 2021-04-30
Fixed bug where GloFAS files are not closed

turnerm:
I've been spoiled with Pandas and didn't realize that `xarray` was not automatically closing files. To fix this:
- converted `open_dataset` to `load_dataset`
- wrapped `open_mfdataset` in a with-as statement

@hannahker perhaps this will solve your Google Drive issue?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ coloredlogs==14.0
 rioxarray==0.1.1
 tqdm==4.50.2
 rasterstats==0.14.0
+cdsapi==0.5.1
 cdstoolbox_remote==0.0.1
 pygsheets==2.0.3.1
 cftime==1.3.0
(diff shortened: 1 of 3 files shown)


hannahker (approved):
This looks good to me and I think fixes my problem!! 🥳 
Just noticed that some things in other files are a bit out of sync, probably because of the somewhat complex branching that's gone on. Eg. reference to ```get_glofas_data_v2.py``` but that is now ```get_glofas_data_old_version.py```. I'm assuming this would be sorted by merging ```develop``` into this branch?

turnerm:
@hannahker so glad that it fixes your problem!

In a previous PR I renamed `get_glofas_data_v2.py` to `get_glofas_data_old_version.py`, because as you pointed out, having `v2` in the name makes it sounds newer than the un-versioned one

PR #65, opened 2021-05-06
New GloFAS model -- bangladesh

turnerm:
- analysis of new GloFAS model and trigger selection
- modify skill figures to use thresholds instead of rain / dry distinction


analyses/bangladesh/trigger_development/read_in_data.py
```diff
@@ -13,11 +13,16 @@
...
 
+pd.options.mode.chained_assignment = None
+
 DATA_DIR = Path(os.environ["AA_DATA_DIR"])
 GLOFAS_DIR = DATA_DIR / "processed/bangladesh/GLOFAS_Data"
 STATION = "Bahadurabad_glofas"
 ffwc_dir = DATA_DIR / 'exploration/bangladesh/FFWC_Data'
 
+# Event definition
+EVENT_WATER_THRESH = 19.5 + 0.85
```
hannahker, line 24: Would it make sense for these kinds of parameters to go in a Bangladesh-specific ```config.yml```? Like in the case of mwi, for example?

analyses/bangladesh/trigger_development/bgd-glofas-new-model.md
```diff
@@ -0,0 +1,404 @@
...
+rp_dict = {}
+for version in [2,3]:
+    f_rp = get_return_period_function(da_glofas_reanalysis[version])
+    print(f'Version {version}')
+    for year in [1.5, 2, 3, 4, 5, 10, 20]:
+        val = 5000*np.round(f_rp(year) / 5000)
+        print(year, val)
+        if version == MAIN_VERSION:
+            rp_dict[year] = val
+# Make 1 in 5 year always 100,000
```
hannahker, line 79: How come these values are being manually set?

analyses/bangladesh/trigger_development/bgd-glofas-new-model.md
```diff
@@ -0,0 +1,404 @@
...
+                ax2.plot(x[a:b], y3[a:b], '-c', lw=3, alpha=0.5)
+                ax2.plot(x[detection], y3[detection], '^', c='m')
+        iax += 1
+
+
+plot_years(df_final, rp_dict[5])
+
+```
+
+### Forecasts
```
hannahker, line 301: I'm not really following these graphs. What is 'Number'?

hannahker (approved):
Looks good to me. Some similar questions as the previous PR about some of the plots, but didn't notice any notable changes to functionality that I'd recommend.

analyses/bangladesh/trigger_development/read_in_data.py
turnerm: Yes definitely. When I change the GloFAS scripts to top-level, I will also make this change

analyses/bangladesh/trigger_development/bgd-glofas-new-model.md
turnerm: For the new model, the 1 in 5 y RP is around 97000, so rounds down to 95000. But since it's so close to the model version 2 1 in 5 year value of 98000, we just round them both up to 100,000. I've added an explantion.

analyses/bangladesh/trigger_development/bgd-glofas-new-model.md
turnerm: Ah sorry these figures are indeed unclear, I've now added an explantion

PR #67, opened 2021-05-07
ecmwf processing and skill

Tinkaa:
-  Retrieving ECMWF's seasonal forecast with monthly precipitation (`ecmwf_seasonal.py` and `get_ecmwf_seasonal_data.py`)
- Convert the dataset to have all forecasted values for a specific month from different lead times (`read_in_data.py`)
- Get started on skill analysis of MWI (`mwi_ecmwf_monthly_skill.md`)
- Compute values per admin1 instead of raster cell (`read_in_data.py`)

@turnerm I tried to follow "your" GloFas structure as much as possible. Would you be willing to have a look if this code is decent?


src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+DATA_DIR = Path(os.environ["AA_DATA_DIR"])
+RAW_DATA_DIR = DATA_DIR / "raw"
+PROCESSED_DATA_DIR = DATA_DIR / "processed"
+ECMWF_SEASONAL_DIR = Path("ecmwf_seasonal_data")
+CDSAPI_CLIENT = cdsapi.Client()
+DEFAULT_VERSION = 5
+
+logger = logging.getLogger(__name__)
+
+class EcmwfSeasonal:
```
turnerm, line 26: There is a lot of repetition between this class and the base GloFAS class -- some of the methods are identical. I think you could probably create a more general base class (maybe a CDS class?), which could go in utils_general or something, and then in flooding / drought we could have the respective glofas / ecmwf subclasses

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+            dataset="monthly_mean",
+            dataset_variable_name="product_type",
+        )
+
+    def download(
+        self,
+        country_name: str,
+        country_iso3: str,
+        area: Area,
+        # leadtimes: List[int],
```
turnerm, line 228: Remove if not used

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+                    year=year,
+                    month=month,
+                    version=version,
+                )
+
+    def process(
+        self,
+        country_name: str,
+        country_iso3: str,
+        # leadtimes: List[int],
```
turnerm, line 253: Remove if not used

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+
+        # Write out the new dataset to a file
+        self._write_to_processed_file(
+            country_name=country_name,
+            country_iso3=country_iso3,
+            ds=ds,
+            version=version,
+        )
+
+#TODO: understand if also the case for ecmwf (copied from glofas)
```
turnerm, line 282: You can test this by just calling the xarray expand dims and see if it crashes

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+            "variable": "total_precipitation",
+            'originating_centre': 'ecmwf',
+            'system': version,
+            "format": "grib",
+            self.dataset_variable_name: self.dataset,
+            "year": str(year),
+            "month": [str(x).zfill(2) for x in range(1, 13)]
+            if month is None
+            else str(month).zfill(2),
+            "leadtime_month":[str(x) for x in range(1,7)]
```
turnerm, line 135: why 7?

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+        return query
+
+    @staticmethod
+    def _read_in_monthly_mean_dataset(filepath_list: List[Path]):
+        """
+        Read in the dataset for each date and combine them
+        """
+
+        def _preprocess_monthly_mean_dataset(ds):
+            dsnew = ds.copy()
```
turnerm, line 151: Why do you need to make a copy?

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+            # step is in timedelta (in nanoseconds), where the timedelta is the end of the valid time of the forecast
+            # since the nanoseconds depends on the length of the month, convert this to the leadtime in months instead to be able to compare across months
+            #other option could be to convert it to the forecasted time, but gets difficult to concat all different publication dates afterwards
+
+            #todo: make this an input variable!
+            dsnew["step"] = range(1,7)
```
turnerm, line 157: Why 7? If it is a parameter of the dataset then you should get it from the dims, otherwise make it a constant

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -0,0 +1,301 @@
...
+        return xr.load_dataset(filepath)
+
+class EcmwfSeasonalForecast(EcmwfSeasonal):
+    def __init__(self):
+        super().__init__(
+            year_min=2000,
+            #TODO: include way to not make it crash with current year (2021)
+            year_max=2020,
+            cds_name="seasonal-monthly-single-levels",
+            dataset="monthly_mean",
```
turnerm, line 219: The dataset variable expects a list type

src/malawi/get_ecmwf_seasonal_data.py
```diff
@@ -0,0 +1,52 @@
...
+import os
+import sys
+
+path_mod = f"{Path(os.path.dirname(os.path.realpath(__file__))).parents[1]}/"
+sys.path.append(path_mod)
+from src.indicators.drought import ecmwf_seasonal
+from src.indicators.flooding.glofas.area import AreaFromShape
+from src.indicators.drought.config import Config
+
+# Location of stations on the Jamuna/Brahmaputra river from http://www.ffwc.gov.bd/index.php/googlemap?id=20
```
turnerm, line 19: Remove comment

src/malawi/get_ecmwf_seasonal_data.py
```diff
@@ -0,0 +1,52 @@
...
+# Some lat lon indicated by FFWC are not on the river and have been manually moved to the closest pixel on the river
+# Bahadurabad_glofas corresponds to the control point identified here:
+# https://drive.google.com/file/d/1oNaavhzD2u5nZEGcEjmRn944rsQfBzfz/view
+COUNTRY_NAME = "malawi"
```
turnerm, line 23: Get these from the config file too (I know I didn't, but I should have!)

turnerm:
Pausing my review for now -- looks great so far! I think the main thing is that we want to avoid repeating code as much as possible, so I think it's important to consolidate (and possible generalize slightly) any repeated functions in an overarching CDS class. 

I also wrote some tests for the GloFAS class -- you can try to extend those to ECMWF. (Actually I've been trying to improve the tests, I will make a PR and add you as a reviewer!)

On Monday I can look at the notebooks.

analyses/malawi/notebooks/read_in_data.py
```diff
@@ -0,0 +1,184 @@
...
+import xarray as xr
+
+path_mod = f"{Path(os.path.dirname(os.path.realpath(__file__))).parents[2]}/"
+sys.path.append(path_mod)
+
+from src.indicators.drought import ecmwf_seasonal
+from src.malawi import get_ecmwf_seasonal_data as gesd
+from src.indicators.drought.config import Config
+
+#TODO: not sure which functions to include here and which in the ecmwf_seasonal file
```
turnerm, line 19: I originally made this file for Bangladesh when I had multiple notebooks using the same methods to avoid duplication. Now that we are using them for other countries, we should probably generalize them somewhere in `src` (although I'm not sure how much overlap there is between GloFAS and ECMWF methods here)

analyses/malawi/notebooks/read_in_data.py
```diff
@@ -0,0 +1,184 @@
...
+    # compute statistics on level in adm_path for all dates in ds
+    df_list = []
+    for date in ds.time.values:
+        for leadtime in ds.leadtime.values:
+            for number in ds.number.values:
+                df = gpd.read_file(adm_path)[[adm_col,"geometry"]]
+                ds_date = ds.sel(time=date,number=number,leadtime=leadtime)
+
+                df[["mean_cell", "max_cell", "min_cell"]] = pd.DataFrame(
+                    zonal_stats(vectors=df, raster=ds_date.values, affine=raster_transform, nodata=np.nan))[
```
turnerm, line 77: If speed is an issue, I found it can be faster to use rasterio mask, see this issue [here](https://github.com/OCHA-DAP/pa-COVID-model-parameterization/issues/35)

src/malawi/get_ecmwf_seasonal_data.py
Tinkaa: You need to define the country name manually I would say, but the country iso3 yes, should come from the config

src/indicators/drought/ecmwf_seasonal.py
Tinkaa: seems to be working fine for me, so will remove. Didn't dig into this, but since you say it should have been fixed, does it have to do anything with versioning that we should indicate in the requirements?

src/indicators/drought/ecmwf_seasonal.py
Tinkaa: Good point. This are the leadtimes, now set it as a global var

src/indicators/drought/ecmwf_seasonal.py
```diff
@@ -207,26 +197,38 @@ def __init__(self):
...
+                if year<current_date.year:
+                    month_range = range(1, 13)
+                elif year == current_date.year:
+                    #forecast becomes available on the 13th of the month
+                    max_month=current_date.month if current_date.day>=13 else current_date.month-1
+                    month_range = range(1,max_month+1)
+                elif year> current_date.year:
+                    logger.info(f"Cannot download data for {year}, because it is in the future")
+            else:
+                month_range = [None]
```
Tinkaa, line 231: @turnerm do you think this is a neat enough way to account for not all months being available for the current year or would you know a programmatically better way?

src/indicators/drought/ecmwf_seasonal.py
Tinkaa: Yes agreed! Not having the time for it now, but made an issue #83

analyses/malawi/notebooks/mwi_ecmwf_monthly_skill.md
```diff
@@ -0,0 +1,382 @@
...
+
+```python
+da_obs=rioxarray.open_rasterio(chirps_monthly_mwi_path,masked=True)
+#only select the years for which we also identified dry spells
+da_obs=da_obs.sel(time=da_obs.time.dt.year.isin(range(2000,2021)))
+```
+
+```python
+#interpolate forecast data such that it has the same resolution as the observed values
+#not sure if nearest or linear is most suitable here..
```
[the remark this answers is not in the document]
Tinkaa: Yes I think the forecasts are designed such that edges should be preserved, so will use the nearest

analyses/malawi/notebooks/read_in_data.py
Tinkaa: Moved it to `src/indicators/drought/ecmwf_seasonal/processing.py`. Do you think that makes sense?

analyses/malawi/notebooks/read_in_data.py
Tinkaa: Wauw this is great if that speeds up things! Since it is an xarray dataset instead of rasterio it needs some changes, but I think it should be possible. Not the time to look into it now, but will make an issue for it! (also in other scripts we are using `zonal_stats` quite a bit so definitely worth looking into)

Tinkaa:
@turnerm thanks so much for the elaborate review. 
I addressed all the comments and made issues for open questions. 
Can you have another check if it looks decent now? 
The `analyses/malawi/notebooks/mwi_ecmwf_monthly_skill.md` became a lot more elaborate. But only have a look at it if time allows (the `set threshold based on x% probability` is the most relevant section)

src/indicators/drought/ecmwf_seasonal/ecmwf_seasonal.py
```diff
@@ -0,0 +1,272 @@
...
+            cds_name="seasonal-monthly-single-levels",
+            dataset=["monthly_mean"],
+            dataset_variable_name="product_type",
+        )
+
+    def download(
+        self,
+        country_iso3: str,
+        area: Area,
+        leadtimes: List[int] = DEFAULT_LEADTIMES,
```
turnerm, line 209: Default arguments should never be mutable, because if you change the variable somewhere in the method then the default will also be changed. Best practice is to set it to `None`, and then in the method do:
```python
if leadtimes is None:
    leadtimes = DEFAULT_LEADTIMES
```

analyses/malawi/notebooks/mwi_ecmwf_monthly_skill.md
```diff
@@ -0,0 +1,946 @@
...
+
+import rioxarray
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+import pandas as pd
+import xskillscore as xs
+import numpy as np
+
+import seaborn as sns
+from mlxtend.evaluate import confusion_matrix
```
turnerm, line 32: This import should be added to the requirements

analyses/malawi/notebooks/mwi_ecmwf_monthly_skill.md
```diff
@@ -0,0 +1,946 @@
...
+#rainy includes the months to select
+#thresh the thresholds
+subset_dict={"rainy":[12,1,2],"thresh":[170]}
+
+for leadtime in da_lt_interp.leadtime:
+    forecast = da_lt_interp.sel(
+    leadtime=leadtime.values).dropna(dim='time')
+    observations = da_obs.reindex({'time': forecast.time}).precip
+    # For all dates
+    crps = xs.crps_ensemble(observations, forecast,member_dim='number')
```
turnerm, line 142: Since we're calculating the CRPS for many different countries now, I've added a method for calculating it [here](https://github.com/OCHA-DAP/pa-anticipatory-action/pull/82/files?file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.yml#). This script should probably eventually be updated to use it.

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -0,0 +1,210 @@
...
+
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+path_mod = f"{Path(os.path.dirname(os.path.realpath(__file__))).parents[2]}/"
+sys.path.append(path_mod)
+
+from src.indicators.drought.ecmwf_seasonal import ecmwf_seasonal
+from src.malawi import get_ecmwf_seasonal_data as gesd
```
turnerm, line 17: To generalize these methods, you shouldn't use Malawi specific information

analyses/malawi/notebooks/mwi_ecmwf_monthly_skill.md
Tinkaa: This is really nice! Yes definitely. I am gonna leave it like this for now if that is alright and then get back to it later.

Tinkaa:
Discussed with @turnerm that `src` code is ready to be merged (and opened issues on next steps). `mwi_ecmwf_monthly_skill_dryspells.md` will be reviewed later in a separate branch

PR #68, opened 2021-05-07
Update GloFAS tests

turnerm:
- Updated the tests to be on the public download method instead of private query method
- Added optional year parameter to the download method -- am undecided if it's really needed. If I keep it then I should also add it to the process methods.


src/indicators/flooding/glofas/glofas.py
```diff
@@ -303,11 +311,15 @@ def download(
         area: Area,
         leadtimes: List[int],
         version: int = DEFAULT_VERSION,
+        year_min: int = None,
+        year_max: int = None,
     ):
```
Tinkaa, line 316: I would say it is nice to have the year option, as I can see use cases where you are only interested in a couple of years and since the downloading is taking so long

Tinkaa (commented):
It looks great to me! Have to admit that I am very new with tests, so might miss things that are not logical, but with my knowledge it all looks good. 

I do see that two .grib files are being checked in (output from tests?). Did you do that on purpose?

src/indicators/flooding/glofas/glofas.py
turnerm: Okay I will leave it then, thanks for the feedback :)

turnerm:
> I do see that two .grib files are being checked in (output from tests?). Did you do that on purpose?

Oops, that was for some more tests which I never finished and moved to another branch. Removed!

PR #77, opened 2021-05-13
Add data directory structure info to the readme

turnerm:


README.md
```diff
@@ -67,5 +62,29 @@ The indicators and analyses folders contain more specific information on the dat
 ├── requirements.txt
 ├── README
 └── LICENSE
+```
+
+## Data directory
 
-```
\ No newline at end of file
+Sync [this](https://drive.google.com/drive/u/3/folders/1RVpnCUpxHQ-jokV_27xLRqOs6qR_8mqQ)
```
Tinkaa, line 69: This seems to link to the `Data` directory instead of the `Data/public`? For now none of the dirs is openly accessible yet, but we should change this link

Tinkaa (commented):
After changing the link, it is good to go

README.md
turnerm: Good point, I'll update the link!

turnerm: Actually, I realized that if we change the link, we also need to change the instructions to say that the environment variable should go to the directory right above `public`, which complicates the set up a bit for external users. Do you know if there is any way we can provide the link to `Data`, but make sure that nothing in `private` is accessible?

Tinkaa: Hmm again good point.. 
> Do you know if there is any way we can provide the link to Data, but make sure that nothing in private is accessible?
I wouldn't logically expect this to be possible and Googling doesn't show an answer. Would you know any way? 
Else another approach might be to have two env variables, `AA_DATA_DIR_PUBLIC` and `AA_DATA_DIR_PRIVATE`? would require us to change part of the paths again but might cover all use cases (or if we have a more genaralised config we can maybe have an even easier method)

turnerm: Opened issue #80 to tackle this, keeping the text as-is in the meantime.

PR #78, opened 2021-05-13
Bangladesh trigger analysis structure

turnerm:
- updated `analyses/bgd/trigger` for new data directory structure
- cleaned up the directory a bit, numbered the scripts / notebooks, put utils in a separate dir


hannahker (approved):
Looks good to me! So the ```utils.py``` file is under the ```analyses``` dir for now, but the idea is that this would eventually all be within ```src```, right? As is in the Nepal branch.

turnerm:
> Looks good to me! So the utils.py file is under the analyses dir for now, but the idea is that this would eventually all be within src, right? As is in the Nepal branch.

Yes exactly!

PR #79, opened 2021-05-13
Feature/floodscan custom areas

hannahker:
Adding in extra functionality for processing Floodscan data: 
- Can specify a start and end date of analysis (also helpful for testing)
- Can get zonal stats using a custom shapefile area (eg. a pre-processed subset of admin boundaries, a buffer around a point). This preprocessing to get the custom shapefile should be done separately. 

The diff might look unnecessarily big because I also reformatted with black 😬 

Keeping this as a draft right now because I haven't tested it very well. If you could take a look as is though that would be great.


src/indicators/flooding/floodscan/floodscan.py
```diff
@@ -32,46 +32,95 @@
...
+        if custom_path:
+            boundaries_path = custom_path
+
+            # this takes a few hours to compute
+            df = self.compute_stats_per_area(
+                ds, coords_transform, boundaries_path, custom_id_col,
+            )
+            self._write_to_processed_file(country_iso3, adm_level, df, custom_name)
+
+        else:
```
Tinkaa, line 90: I would only set the different `boundaries_path` and `id_col` in the if else statement, and then do the `compute_stats_per_area` and `write_to_processed_file` outside the ifelse since they are using the same arguments in both cases.

Tinkaa (commented):
Bit hard to see details indeed with the Black, but looks good to me (and actually great you "blacked" it)

PR #82, opened 2021-05-24
Nepal GloFAS analysis

turnerm:
- Notebook to do the usual analysis on some Nepal GloFAS stations
- refactored methods used for several countries to a general GloFAS utils
- fixed issue with reforecast interpolation


hannahker (approved):
Also looks all good to me! Some of those plotting functions would probably be useful in ```glofas.utils.py``` but that's a future to-do.

src/indicators/flooding/glofas/utils.py
```diff
@@ -0,0 +1,222 @@
...
+    ranges for the two datasets
+    :param da_observations: GloFAS reanalysis at a particular station
+    :param da_forecast: GloFAS reforecast at a particular station
+    :param leadtime: Leadtime
+    :return: Observations and forecast with overlapping values only
+    """
+    forecast = da_forecast.sel(leadtime=leadtime).dropna(dim="time")
+    observations = da_observations.reindex({"time": forecast.time}).dropna(dim="time")
+    forecast = forecast.reindex({"time": observations.time})
+    return observations, forecast
```
hannahker, line 222: So great to have all these functions here. Thanks for doing this!

PR #87, opened 2021-05-28
SSD floodscan and precipitation

Tinkaa:
Code and documentation for the exploratory analysis on the correlation of floodscan and monthly precipitation in South Sudan. It is a very simple analysis and there are many caveats that should be explored further. But
- `analyses/ssd/scripts/ssd_corr_floodscan_monthlyprecip.R ` processes floodscan and precipitation data and looks at their correlations. 
- `analyses/ssd/scripts/ssd_doc_corr_floodscan_monthlyprecip` summarizes the results

@milanoleonardo would you be willing to have a look at this? 
Mainly a read-through of the `analyses/ssd/scripts/ssd_doc_corr_floodscan_monthlyprecip.html` would be good (download and open in browser to view).


Tinkaa:
In coordination with @milanoleonardo decided to merge without review

PR #90, opened 2021-06-03
Jupyter book experiment

Tinkaa:
Start using Jupyter book for documentation, as it enables to combine ipynb as well as md, in python and R. And so far it is really nice! 

Only doubt for me is that it is producing a lot of files. And we do have to merge it to `develop` in order to display it with github pages. 

@hannahker  just wanted to ask if you have an opinion whether it is worth it to introduce all these extra files?


Tinkaa:
@hannahker I am going to merge this now such that I can share a document with github pages, but possible to revert later

PR #99, opened 2021-06-21
Nepal neighbouring basins correlations

turnerm:
We want to know if a trigger in a major basin is correlated with events in nearby smaller basins.

Note that the return period methods and notebook are here but were reviewed in PR #97, I accidentally merged this PR too early (sorry for the mess!)


analyses/npl/05_station_correlation.md
```diff
@@ -0,0 +1,365 @@
...
+    'Babai': ['Chepang']
+}
+STATIONS_BY_MAJOR_BASIN = {
+    'Koshi': ['Chatara', 'Simle', 'Majhitar', 'Kampughat', 'Rai_goan'],
+    'Karnali': ['Chisapani', 'Asaraghat', 'Dipayal', 'Samajhighat', 'Kusum', 'Chepang'],
+}
+```
+
+```python
+ds_glofas_reanalysis = utils.get_glofas_reanalysis(
```
hannahker, line 42: I've encountered and fixed this on another branch that I'm working on, but I think in ```glofas.py``` the ```_get_processed_filepath()``` method needs to be adjusted to add ```public``` to the filepath. It's odd because I'm not sure how this happened...?

analyses/npl/05_station_correlation.md
```diff
@@ -0,0 +1,365 @@
...
+    events_all[basin] = station_events
+```
+
+```python
+x = np.array([1, 2, 3, 4, 4, 3, 2, 3, 4, 3, 2])
+g = utils.get_groups_above_threshold(x, 3.5)
+
+```
+
+### Plot river discharge vs time for all stations
```
hannahker, line 86: I really like these plots! This might be good to add to utils as a generalized plotting function when we have more time...

analyses/npl/06_return_periods.md
```diff
@@ -0,0 +1,55 @@
...
+years = np.arange(1.5, 30.5, 0.5)
+df_rps_empirical = utils.get_return_periods(ds_glofas_reanalysis, years=years, method="empirical")
+df_rps_analytical = utils.get_return_periods(ds_glofas_reanalysis, years=years, method="analytical", show_plots=True)
+```
+
+```python
+glofas_rp = pd.read_excel(GLOFAS_RP_FILENAME, index_col='rp')
+
+
+for i, station in enumerate(stations):
```
hannahker, line 45: ```stations``` needs to be defined in this notebook.

hannahker (approved):
A couple of small things to tidy up but looks good to merge otherwise! It might be worth documenting the difference between the analytical and empirical return period methods at some point. I'll admit that I'm not very clear on what the difference is and when we would use one over the other.

analyses/npl/05_station_correlation.md
turnerm: I think it keeps getting changed back when I merge in develop or something like that. Hopefully after merging this branch it will be fixed!

analyses/npl/05_station_correlation.md
turnerm: Glad you like them! Yes I can definitely generalize this, especially if I find myself using it for another pilot

analyses/npl/06_return_periods.md
turnerm: Oops, fixed!

turnerm:
> It might be worth documenting the difference between the analytical and empirical return period methods at some point. I'll admit that I'm not very clear on what the difference is and when we would use one over the other.

That's a good point, I can try to add some more info to the docstrings in the upstream branches

PR #100, opened 2021-06-21
Bug: GloFAS utils group finder

turnerm:
For finding consecutive groups, the method should use `>=` instead of `>` when comparing to the minimum duration


src/indicators/flooding/glofas/utils.py
```diff
@@ -184,7 +184,7 @@ def get_groups_above_threshold(observations, threshold, min_duration=1):
     groups = np.where(np.diff(observations > threshold, prepend=False, append=False))[
```
Tinkaa, line 184: Only if not too much effort, but maybe change the function name `get_groups_above_threshold` to `get_groups_ge_threshold` ?

Tinkaa (approved):

PR #101, opened 2021-06-21
Nepal water level events

turnerm:
With the water level data from DHM, we can now define flood events as when the water level exceeds the danger level. This notebook analyzes these events and compares them to GloFAS return period exceedance.

One import changes is in the `get_groups_above_threshold`, I changed it from using `>` to `>=`, so it's really getting the groups above or equal to the threshold. I think this makes sense because I think a trigger is activated when a danger level is reached rather than exceeded. With low-precision data (like the water level data) this actually makes a difference as to what constitutes an event. I don't think this change will affect other / past analyses, @hannahker let me know if you think it does.


src/indicators/flooding/glofas/glofas.py
```diff
@@ -209,7 +209,7 @@ def _get_processed_filepath(
         if leadtime is not None:
             filename += f"_lt{str(leadtime).zfill(2)}d"
         filename += ".nc"
-        return DATA_DIR / PROCESSED_DATA_DIR / country_iso3 / GLOFAS_DIR / filename
+        return DATA_DIR / PUBLIC_DATA_DIR / PROCESSED_DATA_DIR / country_iso3 / GLOFAS_DIR / filename
```
hannahker, line 212: Ah I see you've added this in here so probably don't need to make that change on the correlation branch.

src/indicators/flooding/glofas/utils.py
```diff
@@ -180,10 +201,25 @@ def get_crps(
     return df_crps
 
 
-def get_groups_above_threshold(observations, threshold, min_duration=1):
-    groups = np.where(np.diff(observations > threshold, prepend=False, append=False))[
-        0
-    ].reshape(-1, 2)
+def get_groups_above_threshold(
+    observations: ArrayLike,
```
hannahker, line 205: How would ```ArrayLike``` be different than a standard list/array?

analyses/npl/08_water_level_events.md
```diff
@@ -0,0 +1,429 @@
...
+
+## Make plots for presentation
+
+```python
+def plot_arrow(ax, x, y, c):
+    ax.annotate(" ", 
+                     xy=(x, y+y*0.3),
+                     xytext=(x, y+y*0.5),
+                    arrowprops=dict(facecolor=c, shrink=0.05, headlength=3,
+                               width=1, headwidth=3, lw=0.5))
```
hannahker, line 232: Maybe nice to add a bit of transparency to these arrows to better see when they're overlapping.

src/indicators/flooding/glofas/utils.py
```diff
@@ -180,10 +201,25 @@ def get_crps(
...
+    threshold: float,
+    min_duration: int = 1,
+    additional_condition: ArrayLike = None,
+) -> List:
+    """
+    Get indices where consecutive values are equal to or above a threshold
+    :param observations: The array of values to search for groups (length N)
+    :param threshold: The threshold above which the values must be
+    :param min_duration: The minimum group size (default 1)
+    :param additional_condition: (optional) Any additional condition the values must satisfy
```
hannahker, line 215: Nice!

hannahker (approved):
Looks good to merge to me!

src/indicators/flooding/glofas/glofas.py
turnerm: Yes I should have just made a separate PR for this, I've had to add this line in so many branches ><

src/indicators/flooding/glofas/utils.py
turnerm: Oops I had not understood ArrayLike, I am putting it back to Array (ArrayLike is anything that can be coerced to an array)

analyses/npl/08_water_level_events.md
turnerm: Good idea!

src/indicators/flooding/glofas/utils.py
turnerm: Thanks! We should really do this for every method, oh boy...

PR #104, opened 2021-06-24
BFA next step analysis

Tinkaa:
- Analyze the trigger per admin1 instead across the whole region as an experiment (`bfa_seas_bavg_iriforecast_peradm1.md`)
- Analyze the difference between the below average and above average tercile (`bfa_seas_bavg_iriforecast.md`)
- Compute the skill (RPSS) (`bfa_irirpsskill.md`)
- Compute the historical observed values and return periods (`bfa_seas_bavg_precip_corr_obsfor.md`)

I don't have any specific questions for this PR.


Tinkaa:
@joseepoirier  you reviewed most of this PR without realizing ;) So will merge it now.

PR #105, opened 2021-06-24
Feature/rp comp

Tinkaa:
Place part of the return period functionalities in the `utils_general` folder. 
Such that it can be used by other datasets than GloFas as well.

### DIFF
--- a/src/indicators/flooding/glofas/utils.py
+++ b/src/indicators/flooding/glofas/utils.py
@@ -8,7 +8,7 @@ from scipy.stats import rankdata
 import xskillscore as xs
 
 from src.indicators.flooding.glofas import glofas
-from src.utils_general.statistics import _get_return_period_function_analytical, _get_return_period_function_empirical
+from src.utils_general.statistics import get_return_period_function_analytical, get_return_period_function_empirical
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


src/utils_general/statistics.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np
+import pandas as pd
+from scipy.interpolate import interp1d
+
+from scipy.stats import genextreme as gev
+import matplotlib.pyplot as plt
+
+def _get_return_period_function_analytical(
```
turnerm, line 8: The method names should not begin with an underscore, as that is used to denote private methods (i.e. not used outside their own class / module)

src/indicators/flooding/glofas/utils.py
```diff
@@ -108,7 +106,6 @@ def _convert_dict_to_ds(ds_glofas_dict) -> xr.Dataset:
         .rename({"step": "leadtime"})
     )
 
-
```
turnerm, line 111: Sorry to be pedantic but there should be two blank lines between methods (I only noticed because PyCharm complains!)

turnerm (approved):
Looks great! But if you could please hold off on merging until we've merged the Nepal branch, otherwise it will become a total quagmire

Tinkaa:
@turnerm will leave it to you to merge after you merged Nepal, alright?

turnerm:
> @turnerm will leave it to you to merge after you merged Nepal, alright?

Will do!

PR #106, opened 2021-06-24
BFA ReturnPeriod

Tinkaa:
Sorry the PR's are messy, but just wanted to include the changes to the notebook as well. So only the `bfa_seas_bavg_precip_corr_obsfor.md` file is relevant here, the other two files are already included in #105


analyses/bfa/notebooks/bfa_seas_bavg_precip_corr_obsfor.md
```diff
@@ -61,6 +61,7 @@ sys.path.append(path_mod)
 from src.indicators.drought.config import Config
 
 from src.indicators.drought.iri_rainfallforecast import get_iri_data
+from src.utils_general.statistics import _get_return_period_function_analytical, _get_return_period_function_empirical
```
turnerm, line 64: Commenting here but referring to above lines:
- Line 2: Through jupytext you can choose to not save the metadata, this helps other people avoid kernel confusion when they start up the notebook
- Line 49: Did you add myst_nb to the requirements? I'm getting some version conflicts with the current version of nbconvert required

analyses/bfa/notebooks/bfa_seas_bavg_precip_corr_obsfor.md
```diff
@@ -527,47 +528,79 @@ If we want to continue understanding the suitability of this trigger, we therefo
...
+    else:
+        logger.error(f"{method} is not a valid keyword for method")
+        return None
+    df_rps["rp"] = np.round(f_rp(years))
+    return df_rps
+```
+
+```{code-cell} ipython3
+years = np.arange(1.5, 9, 0.5)
+df_rps_analytical = get_return_periods(df_stats_reg, rp_var="bavg_cell",years=years, method="analytical", show_plots=True)
```
turnerm, line 566: Ah, do I understand correctly that values you are fitting the GEV to are percentages? In that case, the distribution has a known maximum (100%) and so the GEV wouldn't really work -- since it's a PDF, the integral should be 1. 

I know that this would mean not being able to compare regions, but did you think about using total area instead of percentage area (or even better, total area x population in that area?) Just because percentage area can lose a bit of meaning if the total area sizes are different and the population density is variable

turnerm (approved):
Looks good! As I said in a comment, I don't think the GEV can really be fitted to a distribution of percentages since there is a cap at 100%.

analyses/bfa/notebooks/bfa_seas_bavg_precip_corr_obsfor.md
Tinkaa: - Hmm I had to save the metadata cause I was experimenting with rendering it with jupyter book. Will at least remove the kernelspec now. 
- Good point, I will add it. Though this is really experimentation, not sure how clean we want to keep the requirements file.

analyses/bfa/notebooks/bfa_seas_bavg_precip_corr_obsfor.md
Tinkaa: Very smart point :) 
It is only one region in this case. And the way it is computed it is hard to get the total area instead (or not worth the effort I would say). So will stick to empirical for now but this is good insight, thanks.

analyses/bfa/notebooks/bfa_seas_bavg_precip_corr_obsfor.md
turnerm: > not sure how clean we want to keep the requirements file

Yeah, that is something that has been on my mind. Maybe we should make a separate requirements file for the analysis directory? Something to add to my running list of issues to add to issues...

PR #107, opened 2021-06-25
Mwi chirpsgefs

Tinkaa:
Clean up and generalize work on CHIRPS-GEFS (15day forecast)
- Download and processing script (`src/indicators/drought/chirpsgefs_rainfallforecast.py`)
- Notebook with analyses for MWI (`analyses/malawi/analyses/malawi/notebooks/mwi_chirpsgefs_corr_dryspells.md`)

Since we decided not to use this source, I would suggest not to review the notebook now as it is massive. So only the .py scripts. 
Also doubting, since we are not using the data for a trigger, if we want the .py scripts in src or not? 

This PR is not time sensitive.


src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+Data limitations:
+- No CHIRPS-GEFS data is available from 01-01-2020 till 05-10-2020.
+This data is available from the [older version of the model](https://data.chc.ucsb.edu/products/EWX/data/forecasts/CHIRPS-GEFS_precip/15day/Africa/precip_mean/),
+but our contact at CHC recommended to not use this
+
+Assumptions
+- The grid cell size is small enough to only look at cells with their centre within the region, not those touching
+"""
+
+import pandas as pd
```
turnerm, line 20: Small PEP-8 comment: standard library imports should come before third party ones, see [here](url)

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def compute_raster_stats(
+    ds,
+    raster_transform,
+    gdf,
+    gdf_id_col,
+    ds_thresh_list=[2, 4, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50],
```
turnerm, line 44: You should not use lists (or other mutable types) as default values, because if they're changed in the method itself, then the default value also changes. The standard practice is to set the parameter to `None` and then in the method check `if ds_thresh_list is None` to set the default.

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+from src.utils_general.utils import download_ftp
+
+logger = logging.getLogger(__name__)
+
+
+def compute_raster_stats(
+    ds,
+    raster_transform,
+    gdf,
+    gdf_id_col,
```
turnerm, line 43: This parameter is missing from the docstring

src/indicators/drought/chirpsgefs_rainfallforecast.py
turnerm: Does it make sense to set this as a default? What are the units of these thresholds? Will you often be using the exact same values?

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+            vectors=df, raster=ds.values, affine=raster_transform, nodata=np.nan
+        )
+    )["min"]
+    df.loc[:, "mean_cell"] = pd.DataFrame(
+        zonal_stats(
+            vectors=df, raster=ds.values, affine=raster_transform, nodata=np.nan
+        )
+    )["mean"]
+
+    for thres in ds_thresh_list:
```
turnerm, line 71: Small point, but maybe it would read more easily if you called this variable `thresh`?

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+        day_year = str(date.timetuple().tm_yday).zfill(3)
+        # start_day is the day of the year for which the forecast starts
+        download_ftp(
+            config.CHIRPSGEFS_FTP_URL_AFRICA.format(
+                days_ahead=str(days_ahead).zfill(2), year=year, start_day=day_year
+            ),
+            chirpsgefs_filepath,
+        )
+        logger.debug(f"...successfully downloaded {chirpsgefs_filepath}")
+    except Exception as e:
```
turnerm, line 225: It's better to use the specific exception, I think here would be `urllib.error.HTTPError`. You could even check the error string and check that it matches the "HTTP Error 404", so that you don't accidentally miss other http errors

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+        all_dates = all_dates.union(seas_range)
+    return all_dates
+
+
+def download_chirpsgefs(date, config, days_ahead, use_cache=True):
+    """
+    Download the chirps-gefs africa forecast for the given year and day of the year
+    We are focussing on the Africa data, since global data gets massive. Nevertheless, even for Africa it gets massive.
+    Note that a part of the data for 2020 is missing due to a change in model
+    days: number of days forecast predicts ahead. When using 15 day, set to empty string
```
turnerm, line 200: I think this docstring is outdated?

src/indicators/drought/config.py
```diff
@@ -16,107 +17,138 @@ class Config:
...
+    CHIRPS_SEASONAL_LOWERTERCILE_COUNTRY_FILENAME = (
+        "{country_iso3}_chirps_seasonal_lowertercile.nc"
+    )
+    CHIRPS_LON = "longitude"  # "x" #
+    CHIRPS_LAT = "latitude"  # "y" #
     CHIRPS_VARNAME = "precip"
 
-
+    ### CHIRPS-GEFS
+    CHIRPSGEFS_DIR = "chirpsgefs"
```
turnerm, line 149: Should this maybe be "chirps_gefs"? That's the name of the dir in `Data/public/raw/glb/`

src/malawi/get_chirpsgefs_data.py
```diff
@@ -0,0 +1,57 @@
...
+DRY_SPELLS_DIR = "dry_spells"
+RAINY_SEASON_PATH = (
+    COUNTRY_DATA_PROCESSED_DIR
+    / DRY_SPELLS_DIR
+    / "rainy_seasons_detail_2000_2020_mean_back.csv"
+)
+CHIRPSGEFS_PROCESSED_DIR = COUNTRY_DATA_PROCESSED_DIR / CONFIG.CHIRPSGEFS_DIR
+
+# how many days ahead the forecast predicts. Can be 5, 10, or 15
+DAYS_AHEAD = 15
```
turnerm, line 36: Should this maybe be a variable to loop over? I noticced that in the "chirpsgefs" directory, there is 5-day rather than 15-day

src/malawi/get_chirpsgefs_data.py
```diff
@@ -0,0 +1,57 @@
...
+ADM_LEVEL = 2
+
+
+def main(download=True, process=True):
+    if download:
+        rainy_dates = get_rainy_season_dates(RAINY_SEASON_PATH)
+        for d in rainy_dates:
+            download_chirpsgefs(pd.to_datetime(d), CONFIG, DAYS_AHEAD)
```
turnerm, line 44: When I try running this, I still end up downloading files, even though I would expect them all to be cached -- do you know why this is?

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+
+    date_str = date.strftime("%Y%m%d")
```
turnerm, line 203: The dates in the filenames would be a bit more readable with hyphens (but don't worry about changing this now!)

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+import os
+
+path_mod = f"{Path(os.path.dirname(os.path.realpath(__file__))).parents[2]}/"
+sys.path.append(path_mod)
+from src.utils_general.utils import download_ftp
+
+logger = logging.getLogger(__name__)
+
+
+def compute_raster_stats(
```
turnerm, line 39: It would help to have the methods in this file in the order that they are used, e.g. first the download method, then `compute_stats_rainyseason` and finally the methods that it calls (which could be made private by adding `_` to their names)

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def get_rainy_season_dates(
+    rainy_season_path, onset_col="onset_date", cessation_col="cessation_date"
+):
+    """
+    Compute a datetimeindex with all dates across all rainy seasons.
+    This is used to only download data for these dates, since the files are large
+    Args:
+        rainy_season_path: filepath to csv with end and start dates of the rainy season, possibly per admin
```
turnerm, line 159: missing a couple of parameters

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    df_rain = pd.read_csv(rainy_season_path, parse_dates=[onset_col, cessation_col])
+
+    # remove entries where there is no onset and no cessation date, i.e. no rainy season
+    df_rain = df_rain[
+        (df_rain.onset_date.notnull()) | (df_rain.cessation_date.notnull())
+    ]
+    # if onset date or cessation date is missing, set it to Nov 1/Jul1 to make sure all data of that year is downloaded. This happens if e.g. the rainy season hasn't ended yet
+    df_rain[df_rain.onset_date.isnull()] = df_rain[df_rain.onset_date.isnull()].assign(
+        onset_date=lambda df: pd.to_datetime(f"{df.season_approx.values[0]}-11-01")
```
turnerm, line 169: Maybe make Nov 1 and Jul 1 constants

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+    Args:
+        rainy_season_path: filepath to csv with end and start dates of the rainy season, possibly per admin
+    """
+    df_rain = pd.read_csv(rainy_season_path, parse_dates=[onset_col, cessation_col])
+
+    # remove entries where there is no onset and no cessation date, i.e. no rainy season
+    df_rain = df_rain[
+        (df_rain.onset_date.notnull()) | (df_rain.cessation_date.notnull())
+    ]
+    # if onset date or cessation date is missing, set it to Nov 1/Jul1 to make sure all data of that year is downloaded. This happens if e.g. the rainy season hasn't ended yet
```
turnerm, line 167: line too long

src/indicators/drought/chirpsgefs_rainfallforecast.py
turnerm: Also, the date format should be made a constant, since it's used in multiple places

src/indicators/drought/chirpsgefs_rainfallforecast.py
```diff
@@ -0,0 +1,227 @@
...
+                )
+            else:
+                logger.debug(f"Retrieving stats for {date_str}")
+                rds = rioxarray.open_rasterio(chirpsgefs_raster_filepath)
+                df = compute_raster_stats(
+                    rds.sel(band=1), rds.rio.transform(), gdf_adm, adm_col
+                )
+                df.loc[:, "date"] = d
+                df.loc[:, "date_forec_end"] = df.loc[:, "date"] + timedelta(
+                    days=days_ahead - 1
```
turnerm, line 147: Why -1 here?

turnerm (approved):
Looks good! I wasn't able to run the processing however, because the required data is not at the expected path. Let me know once it's there and I can give it a try.

src/indicators/drought/chirpsgefs_rainfallforecast.py
Tinkaa: A part of the link seems to be missing :(  https://github.com/OCHA-DAP/pa-anticipatory-action/pull/url

src/indicators/drought/chirpsgefs_rainfallforecast.py
Tinkaa: Actually you are right, this shouldn't be a default at all as probably use different values for different purposes. The units are mm

src/malawi/get_chirpsgefs_data.py
Tinkaa: Aha so for some of the `rainy_dates` no data is available (e.g. in 1999). So at least I would expect to show the "downloading" message and therafter the message for the same date that the data is not available. Is that correct? I now removed the "downloading" message, as the files are small so downloading only takes a short time. Hope that makes it less messy.

src/indicators/drought/config.py
Tinkaa: `chirps_gefs` is the old directory and is deleted on my local version as well as on the GDrive online. Is the dir for you still visible?

Tinkaa:
Thanks so much! 
Regarding the processing, for me it seems to work.. Could it be something with the syncing of `chirpsgefs` and `chirps_gefs`? Or the cached versions? Aah or are there dates showing up that are not found, but is this because the raw data doesn't exist for these dates? 🧐 I am guessing the last. Not sure how to make that error message clearer..

src/indicators/drought/chirpsgefs_rainfallforecast.py
Tinkaa: cause the `days_ahead` is the number of days the forecast is valid for. The first date of the forecast is extracted from the filename, so date `days_ahead-1` ahead - current date = days_ahead

src/indicators/drought/chirpsgefs_rainfallforecast.py
Tinkaa: nvm, now understood what you meant after reading [the documentation](https://www.python.org/dev/peps/pep-0008/#imports), thanks!

Tinkaa:
@turnerm to me it seems alright to merge now. Let me know if the two folders are still showing up for you!

src/indicators/drought/chirpsgefs_rainfallforecast.py
turnerm: Oops yes that is the link I meant to add, sorry about that ;)

src/indicators/drought/config.py
turnerm: Ah you are right, that is so weird. I deleted my local version, let me know if this causes you any issues...

src/malawi/get_chirpsgefs_data.py
turnerm: No, I meant that after it gets through the no dates files, it starts downloading new ones

PR #110, opened 2021-06-28
MWI update dry spell doc

Tinkaa:
Around the beginning of the pilot we compiled a document describing the rainy seasons and dry spells according to our analysis with CHIRPS. 

We changed the definition in the meanwhile and never finished that document with it. This PR updates the document, and removes the document with the old definition. 

No specific questions. Mostly copied from the old document, and made the numbers variable.


hannahker (approved):
Kudos for updating the documentation 🎉 A couple of small things but otherwise looks good to merge to me! 

The text above this plot wraps a bit weirdly with some of it coming over to the left side: 

Capture

The writeup in the 'Duration' section also has some NA references that I'm assuming aren't intentional? 

> Dry spells varied in duration from 14 days (by definition) to NA days. Overall districts experienced dry spells that lasted an average of NA days; individual districts’ means ranged from NA to NA days, confirming geographical variability in duration as well as in frequency.

Tinkaa:
Thanks @hannahker !

PR #113, opened 2021-06-29
produce world map of AA countries

joseepoirier:
Manually created the entry for Malawi to simplify the graph. We'll need to adapt the code when we have > 1 dual shock context. 
I'd vote for continuing to distinguish between dry spells and drought, or we combine them and label the category drought/dry spells since they are distinct shocks.


utils_general/aa-countries.R
```diff
@@ -0,0 +1,75 @@
+###
```
Tinkaa, line 1: Not regarding this line, but the directory structure. I would propose to not place this in utils_general as this are really pipeline scripts. Instead I would propose to make a `analyses/glb` folder, and place it there. What do you think?

utils_general/aa-countries.R
```diff
@@ -0,0 +1,75 @@
...
+# Install packages
+###
+
+library(tidyverse)
+#library(rvest)
+#library(magrittr)
+#library(ggmap)
+#library(stringr)
```
Tinkaa, line 9: remove unused libraries

utils_general/aa-countries.R
```diff
@@ -0,0 +1,75 @@
...
+world_map <- map_data("world")
+
+###
+# identify AA pilots
+###
+
+# list countries and shocks
+drought <- c("Somalia", "Ethiopia", "Niger", "Burkina Faso")
+#dry_spells <- c("Malawi")
+#floods <- c("Bangladesh", "Nepal", "Malawi", "Chad", "South Sudan")
```
Tinkaa, line 24: remove commented lines

utils_general/aa-countries.R
```diff
@@ -0,0 +1,75 @@
...
+       ,subtitle = "Active and under development frameworks") +
+  theme(legend.position="bottom"
+        #text = element_text(family = "Gill Sans", color = "#FFFFFF")
+        ,plot.title = element_text(size = 24)
+        ,plot.subtitle = element_text(size = 12)
+        ,axis.text = element_blank()
+        ,axis.title = element_blank()
+        ,axis.ticks = element_blank()
+        ,panel.background = element_rect(fill = "white")
+        #,plot.background = element_rect(fill = "#444444")
```
Tinkaa, line 72: here and above, remove commented lines

utils_general/aa-countries.R
```diff
@@ -0,0 +1,75 @@
...
+        ,panel.grid = element_blank()
+  )
+
```
Tinkaa, line 75: don't see code to save the png? (do comment out that line but good to keep it somewhere)

Tinkaa (commented):
Beautiful how ggplot can so easily create nice plots, and you nicely documented the code! 

Few things
- as commented would place the files in a different directory
- I am not a big fan of putting the pngs in the repo. Would it be alright to put it on the GDrive (`processed/glb/plots` for example) or is there a specific reason you saved it here? 
- I think it would be really nice to add the map to the repo main readme. What do you say?

utils_general/aa-countries.R
joseepoirier: Oh I hadn't noticed glb. Much better location for it.

utils_general/aa-countries.R
joseepoirier: Ha! forgot to do that. Thanks!

utils_general/aa-countries.R
joseepoirier: I need to figure out how to move the Rproj file with it..

utils_general/aa-countries.R
joseepoirier: I was leaving them in so we can easily update it / change how we represent MWI. I guess if the overall structure & pattern are clear I can remove them.

utils_general/aa-countries.R
Tinkaa: `glb` doesn't exist yet😁 but think this is a perfect occasion to create it

utils_general/aa-countries.R
Tinkaa: to me the overall structure is clear

joseepoirier:
@Tinkaa I addressed your helpful comments, thanks. Can you take a quick look so we can merge?

Tinkaa:
@joseepoirier Looks great! The only thing is that when I open the png now saved in `processed/glb/plots`, it doesn't look good, it is very stretched. So maybe change how you save the png? But regardless of that, ready to be merged imo.

Tinkaa:
Aah and the png seems to still be saved on the github, would you agree to remove it from there?

utils_general/aa-countries.R
joseepoirier: The saving function distorts the image when it saves it. I've already spent a few hours trying to figure it out. Would say to leave it as, keeping the manually saved png for now, and I'll look into the saving issue later.

joseepoirier:
> Aah and the png seems to still be saved on the github, would you agree to remove it from there?

Ugh that's me forgetting to actually removing it after realising the saving function issue. Sorry - I deleted it and also moved the R file + Rproj to `analyses/glb`

PR #115, opened 2021-07-02
Fix bad merge

turnerm:
I messed up a merge and left the HEAD line in one of the GloFAS files, this PR is to remove it


hannahker (approved):
Good catch 👍

PR #116, opened 2021-07-02
Nepal update station coords

turnerm:
Update station coordinates to GloFAS model v3


hannahker (approved):
Looks good! Aside from checking the differences in values, would there be any reason to keep the v2 station locations?

turnerm:
> Aside from checking the differences in values, would there be any reason to keep the v2 station locations?

I was going to keep them for now because there are a lot of older slide decks with figures created using v2

PR #118, opened 2021-07-05
MWI restructure folders

Tinkaa:
This PR builds on the branch `mwi-cleanup`, in which already a part of the restructuring was done. 

In addition this branch does the following: 
- add scripts and notebooks to one folder named `dryspells_trigger`
- number scripts and notebooks that are part of the main analysis. Proposing this structure to not have to remove files that might be relevant in the future, but do have a clearer impression which files are the main parts of the analysis. 
- Clean-up and add more text to `02_mwi_chirps_observed_monthly_seasonal_precipitation.md` and `03_mwi_corr_observed_belowaverage_precipitation_dryspells.md`
- Clean-up `04_mwi_corr_observed_monthlytotalrainfall_dryspells.md` and merge relevant parts of `mwi_monthlyseasonaltotal_corr_dryspells_adm2.ipynb` into it

Imo this doesn't need a real code review, mainly a check if we agree with the proposed structure.
Note that soon the flooding work will be added in a folder named `flooding_trigger`


hannahker (approved):
Looks good! My only main suggestion would be to remove the ```mwi``` prefix in file names since I think it's a bit redundant as all files are in the Malawi directory. 

And to confirm my understanding of your use of numbering: The idea is that all files prefixed with a number (01, 02, etc.) are considered the main components of an analysis and all other files are more exploratory work? If so, maybe we would want to have an ```exploration``` subdirectory where these are kept (so that the main trigger directory isn't too cluttered)? I could implement this for floods too. So it would be: 

```
 ├── dryspells_trigger
 |           ├── 01_analysisA.md
 |           ├── 02_analysisB.md
 |           ├── exploration
 |           |          ├── exploration1.R

etc...

```

Tinkaa:
Thanks for reviewing! 

> My only main suggestion would be to remove the mwi prefix in file names since I think it's a bit redundant as all files are in the Malawi directory.

Yes I was doubting about this, the reason I did leave it was because if doing the same analysis for another country, it is really nice to be sure you are editing the correct file. But maybe this is abundant. Removing it for now, and adding it to the coding practices doc

> The idea is that all files prefixed with a number (01, 02, etc.) are considered the main components of an analysis and all other files are more exploratory work?

Yes and no😁 part of them are, but part are also scripts with functions being called in the main components. And then also the script to plot heatmaps, which is not really a main component but also not so much exploration. 
Struggling to find a right way to structure this, cause I agree it clutters the dir. For now moving all of it to `exploration` and will think if there is a better structure.

Tinkaa:
@joseepoirier I will merge this based on Hannah's comments, but please let me know if there is anything you don't agree with!

PR #119, opened 2021-07-05
Feautre/pre commit

Tinkaa:
Implementing pre-commit hooks for black and flake8. 
Was looking into it, and seemed to me flake8 would be pretty nice in addition to black.


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+ignore = E203, E266, E501, W503, F403, F401
+max-line-length = 79
+max-complexity = 18
+select = B,C,E,F,W,T4,B9
```
turnerm, line 5: I think some of these recommendations are outdated. According to [this article](https://mitelman.engineering/blog/python-best-practice/automating-python-best-practices-for-a-new-project/) you really only need `extend-ignore = E203`. You can keep the `max-line-length` one if you also set the black one to be 79.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+#largely copied from https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/
+repos:
+    - repo: https://github.com/pre-commit/pre-commit-hooks
+      rev: v2.4.0
+      hooks:
+          - id: flake8
```
turnerm, line 6: I would run Black before flake8, otherwise flake8 will catch a bunch of errors missed by Black

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
...
+    - repo: https://github.com/psf/black
+      rev: stable
+      hooks:
+          - id: black
```
turnerm, line 10: [This article](https://mitelman.engineering/blog/python-best-practice/automating-python-best-practices-for-a-new-project/) also recommends the following config file for Black (you put it in a file called `pyproject.toml`): 

```
[tool.black]
line-length = 79
target-version = ['py38']
include = '\.pyi?$'
exclude = '''

(
  /(
      \.eggs         # exclude a few common directories in the
    | \.git          # root of the project
    | \.hg
    | \.mypy_cache
    | \.tox
    | \.venv
    | _build
    | buck-out
    | build
    | dist
  )/
  | foo.py           # also separately exclude a file named foo.py in
                     # the root of the project
```
But as they say, really the most important line is `line-length`, it seems to work fine without the rest. (And the reason we want to use 79 is because of PEP-8)

turnerm:
Looks good, but the last step is to actually run it and then try to fix all of the flake8 complaints :stuck_out_tongue_winking_eye:. You can run it with the command `pre-commit run --al-files`.

Tinkaa:
Thanks for the tips! 
The pre-commit is passing now, but not sure how to properly review this😅 
But I added some `# noqa` 's , meaning flake8 should ignore those lines. Maybe it is worth checking if you would agree? 
And could you check the change I made on line 164 of `analyses/bgd/pilot_evaluation/scripts/Generate_interpolated.py`?

src/indicators/drought/iri_rainfallforecast.py
```diff
@@ -4,89 +4,117 @@
 from pathlib import Path
 import rasterio
 import xarray as xr
-import rioxarray
+import rioxarray  # noqa: F401
```
turnerm, line 7: What happens here if you don't import rioxarray?

src/indicators/flooding/floodscan/floodscan.py
```diff
@@ -3,7 +3,10 @@
 import geopandas as gpd
 from rasterstats import zonal_stats
 import xarray as xr
-import rioxarray
+
+# not using rioxarray directly but using .rio
+# so let flake8 ignore it
+import rioxarray  # noqa
```
turnerm, line 9: Why did you use `# noqa: F401` in the other file and not here?

turnerm (approved):
Looks great, thanks so much for doing this Tinka!

src/indicators/flooding/floodscan/floodscan.py
Tinkaa: Laziness😇 changed now

Tinkaa:
@turnerm thanks so much! I solved the merge conflicts. Let me know if you think there is anything we can do about the `rioxarray` thing, else I will merge it

src/indicators/drought/iri_rainfallforecast.py
Tinkaa: As discussed, this was an error made up in my head, now removed the unused rioxarray imports

PR #120, opened 2021-07-05
Bugfix: Updated GloFAS API to round to .x5

turnerm:
For downloading GloFAS data from the CDS API, you need to pass the area rounded as .x5. This PR corrects the issue, but also allows a legacy parameter for accessing the incorrect data for now. Should probably be removed at a later date. 

@hannahker what do you think about the legacy directory structure? Right now I just tack "incorrect_coords" on to the version directory. Do you think it's better to put a separate directory in the version directory instead or something? What about for the processed data? (I just changed the filenames there).

Note that you will need to change the directory & filenames by hand. I've done so for Nepal, will take care of Bangladesh at a later date.

### DIFF
--- a/analyses/npl/10_glofas_coord_check.md
+++ b/analyses/npl/10_glofas_coord_check.md
@@ -1,0 +1,87 @@
+Check that the positions of the GloFAS reporting points make sense on the raster
+
+```python
+from pathlib import Path
+import os
+import sys
+import glob
+
+import pandas as pd
+import xarray as xr
+import numpy as np
+import matplotlib.pyplot as plt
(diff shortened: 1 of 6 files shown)


hannahker (approved):
Nice quick fix! This looks good to me. 

Adding in the ```use_incorrect_area_coords``` makes sense right now while we're testing the impact that this has had on past data, but I think we should probably remove it at some point as I can't imagine another use case and it might be confusing to others. 

Also happy with your proposed naming/directory structure for the old files. Since we're not likely to use this incorrect data much it's probably not worth overthinking either 😄

PR #121, opened 2021-07-08
GloFAS reforecast by year

turnerm:
Update GloFAS forecast / reforecast to not be split by leadtime but by year only. I wanted to do this for awhile but now it's necessary due to the data the Chris gave us where the leadtimes are combined.


hannahker (approved):
This looks good to me - thanks for the quick fix here and I like the general refactoring that you've done with the forecast base class.

PR #122, opened 2021-07-09
Feature/mwi docs

Tinkaa:
Create a Rmarkdown/html document summarizing all the analyses related to correlating indicators with observed dry spells. 

- This document can be found in `analyses/malawi/docs/mwi_technical_background_pilot.Rmd`. I think the easiest review is to open the html page, and mark any things you think should be improved. This is more about the text than the code
- All other changed files are just small adjustments in e.g. filenames of saved figures that are used in the doc. No need to really review

With this, the mwi dry spell work cleaning up is also finished in my opinion, so if there is anything else, let me know.


Tinkaa:
The documentation has extensively been discussed with @joseepoirier bilaterally. Merging this now.

PR #123, opened 2021-07-13
Aa viz

joseepoirier:
Sorry I thought I had PRed this already. The basic code to produce the map of all a countries with AA pilots, including the fix to save the png programmatically.


Tinkaa:
Think you PRed everything except the saving before, and that looks good so will merge it :)

PR #124, opened 2021-07-13
initial commit create visualizations for MWI

joseepoirier:


analyses/malawi/dryspells_trigger/data-viz.R
```diff
@@ -0,0 +1,70 @@
...
+options(scipen = 999)
+
+# set directory paths
+# AA_DATA_DIR is set as a variable in .Renviron or .bashprofile
+data_dir <- Sys.getenv("AA_DATA_DIR")
+shapefile_path <- paste0(data_dir, "/public/raw/mwi/cod_ab/mwi_adm_nso_20181016_shp")
+plot_path <- paste0(data_dir, "/public/processed/mwi/plots/dry_spells/dry_spell_plots")
+
+# import dry spells groundtruth dataset
+dry_spells_groundtruth_dataset_url <- "https://data.humdata.org/dataset/df68065d-704c-4556-b111-394df000cee4/resource/951c0454-23ab-4057-a7fb-7cfeed4f842d/download/full_list_dry_spells.csv"
```
Tinkaa, line 22: Nice!

analyses/malawi/dryspells_trigger/data-viz.R
```diff
@@ -0,0 +1,70 @@
...
+#####
+
+# summarise dry spells frequency per district / year
+ds_counts_per_year <- ds %>%
+  filter(season_approx != "outside rainy season") %>%
+  group_by(pcode, season_approx) %>%
+  summarise(n_ds = n())
+
+ds_counts <- ds %>%
+  filter(season_approx != "outside rainy season") %>%
```
Tinkaa, line 39: From the map later it seems to me that the count of number of dry spells is really high. I checked in the `01_chirps_dry_spell_detection.R` script how `outside rainy season` is defined, and it is defined as `data_long$season_approx <- ifelse(data_long$month >= 10, data_long$year, ifelse(data_long$month <= 7, data_long$year - 1, 'outside rainy season'))` 

So that is a very loose definition. We should probably change this in the HDX file.. 

I am alright with merging as it is if you are not using the map for anything official now, but would at least add a comment.

analyses/malawi/dryspells_trigger/data-viz.R
```diff
@@ -0,0 +1,70 @@
...
+# combine dry spells groundtruth dataset and shapefiles
+data <- mwi_adm2 %>%
+  left_join(ds_counts, by = c("ADM2_PCODE" = "pcode"))
+
+# create map
+
+data %>%
+  ggplot() +
+  geom_sf(aes(fill = n_ds)) +
+  scale_fill_continuous(type = "viridis", "Total number of events") +
```
Tinkaa, line 52: Minor thing, but wouldn't use `viridis` as a color scale but instead a sequential scale such that it is clear that a darker hue indicates more events.

Tinkaa (commented):
Nice initiative, thanks! 
You were working with an outdated `develop` branch, so merged the newest version into this branch. To do so I moved the `data-viz.R` to the `dryspells_trigger` folder, as we are not working with the `scripts` folder anymore. 

Two things on file structure: 
- There is also an `exploration` folder within the `dryspells_trigger` folder. Up to you, but it might fit better there?  As this script is not really part of the analyses workflow.
- I would give the file a more descriptive name. E.g. `mwi_dry_spells_map.R` 

Apart from that, see the note on the rainy season definition and then it is good to go imo.

analyses/malawi/dryspells_trigger/data-viz.R
joseepoirier: Ah, that's right. I did mean to select the variable that indicates whether the DS was during the (actual) rainy season or not, the season_approx label. The original script for dry_spell_detection had that. Will follow up.

analyses/malawi/dryspells_trigger/data-viz.R
joseepoirier: I think I had started with that and found the hues unclear. I'll try again to confirm.

analyses/malawi/dryspells_trigger/data-viz.R
joseepoirier: The variable `during_rainy_season` uses rainy season onset and cessation dates to determine whether DSs were during the rainy season. Updated the filter ln 39 and ln 45 accordingly.

analyses/malawi/dryspells_trigger/data-viz.R
joseepoirier: It actually looks fine; putting the blue hues in and reversed it so darker hues represent higher counts.

joseepoirier:
Saved the modified plot on GDrive after addressing Tinka's comments.

joseepoirier:
Addressed Tinka's comments. Ready to merge if she accepts the revisions.

Tinkaa:
The script was not working, but assume you used the newer version of `full_list_dry_spells.csv`. Now updated the HDX dataset to have that newer version. Let me know if that is incorrect. 

Also renamed the file and moved it to the `exploration` folder as we discussed. Let me know if you disagree.

PR #125, opened 2021-07-19
generalize mpe

Tinkaa:
moving the function to `utils_general` and taking out the mean calculation such that it can be used more generally.

### DIFF
--- a/analyses/npl/02_glofas_skill.md
+++ b/analyses/npl/02_glofas_skill.md
@@ -16,6 +16,7 @@ path_mod = f"{Path(os.path.dirname(os.path.realpath(''))).parents[0]}/"
 os.chdir(path_mod)
 
 from src.indicators.flooding.glofas import utils, glofas
+from src.utils_general.statistics import calc_mpe
 import src.nepal.get_glofas_data as ggd
 
 reload(utils)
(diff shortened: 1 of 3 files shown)


src/utils_general/statistics.py
```diff
@@ -92,3 +92,12 @@ def get_return_period_function_empirical(df_rp: pd.DataFrame, rp_var: str):
     df_rp["exceedance_probability"] = df_rp["rank"] / (n + 1)
     df_rp["rp"] = 1 / df_rp["exceedance_probability"]
     return interp1d(df_rp["rp"], df_rp[rp_var])
+
+
+def calc_mpe(observations: np.array, forecast: np.array) -> float:
+    denominator = observations
```
turnerm, line 98: I think this line can be removed and `observations` can be added directly as the denominator.

src/utils_general/statistics.py
```diff
@@ -92,3 +92,12 @@ def get_return_period_function_empirical(df_rp: pd.DataFrame, rp_var: str):
     df_rp["exceedance_probability"] = df_rp["rank"] / (n + 1)
     df_rp["rp"] = 1 / df_rp["exceedance_probability"]
     return interp1d(df_rp["rp"], df_rp[rp_var])
+
+
+def calc_mpe(observations: np.array, forecast: np.array) -> float:
```
turnerm, line 97: Maybe you can add a docstring, like in the functions above?

turnerm: (and btw the other docstring formats are a bit messed up from the Black / flake8 formatting, and I didn't catch it in the PR ><. Each `param` should be on a new line)

src/indicators/flooding/glofas/utils.py
```diff
@@ -324,16 +324,6 @@ def get_rank(observations: np.array, forecast: np.array) -> np.array:
     return rank
 
 
-def calc_mpe(observations: np.array, forecast: np.array) -> float:
```
turnerm, line 327: Removing this completely will break some GloFAS notebooks. Do you think that maybe you could add the method back, and have it call the one from `utils_general.statistics`?

analyses/npl/02_glofas_skill.md
```diff
@@ -164,9 +165,11 @@ for basin, stations in STATIONS_BY_MAJOR_BASIN.items():
         mpe_ev = np.empty(len(da_forecast.leadtime))
         for ilt, leadtime in enumerate(da_forecast.leadtime):
             observations, forecast = utils.get_same_obs_and_forecast(da_observations, da_forecast, leadtime)
-            mpe[ilt] = utils.calc_mpe(observations, forecast)
+            mean_forecast = forecast.mean(axis=0)
+            mpe[ilt] = calc_mpe(observations, mean_forecast)
```
turnerm, line 169: Thanks for updating this!

turnerm (approved):

src/utils_general/statistics.py
Tinkaa: Oh no! good catch, apparently the tool I used to wrap them messed this up. Fixed them here, but didn't feel like fixing all of them 😅

PR #127, opened 2021-07-20
mwi docs - add caveats

Tinkaa:
added a list of caveats to the mwi dry spells dataset docs. 
Thought it made more sense to add them here than only on HDX. We can then also add them to the metadata of the [HDX dataset](https://data.humdata.org/dataset/historical-dry-spells-in-malawi). What do you think?


joseepoirier:
The first goal was to provide a richer description of caveats to HDX users, as they might use the data without understanding how they were generated.
Agreed that they should be added here as well, esp. because we haven't had a chance to refactor the code and document those limitations and decision points in the code.

analyses/malawi/dryspells_trigger/exploration/mwi_viz_dry_spells_map.R
```diff
@@ -46,7 +46,7 @@ data <- mwi_adm2 %>%
   left_join(ds_counts, by = c("ADM2_PCODE" = "pcode"))
 
 # create map
-data %>%
+plt_ds_map <- data %>%
```
joseepoirier, line 49: If we assign the plot to a variable, I don't think the saving function ln 70 will work, since it saves the plot last generated in Viewer. Did you test it?

An option might be to assign it, then call `plt_ds_map` on ln 68 before it gets saved.

Tinkaa:
Yes, this was just to prevent duplicate work. Do you think the caveats are fine?

analyses/malawi/dryspells_trigger/exploration/mwi_viz_dry_spells_map.R
Tinkaa: Uhuh good catch. I tested it and it seems to work, but safer to indicate to ggsave that it is plt_ds_map that it should save, so added that now.

joseepoirier:
I think the caveats are fine. We might want to document why we have "overlapping" dry spells? Otherwise, good to merge.

Tinkaa:
Good point, added and will merge. Also changed the name of the file

PR #128, opened 2021-07-23
refactor crps

Tinkaa:
proposal to generalize the CRPS
- largest change is to do the normalization after selecting on a threshold. lets discuss if that makes sense. 
- I tried to generalize the functions more, but since glofas uses >thresh and ecmwf <=thresh I couldn't really make it work. Let me know if you think there is a better way to generalize. 
- For ecmwf it is really slow. I think it is mainly the `xs.crps_ensemble` so we can probably not change much about that, but let me know if you think we can.

### DIFF
--- a/src/indicators/drought/ecmwf_seasonal/processing.py
+++ b/src/indicators/drought/ecmwf_seasonal/processing.py
@@ -90,13 +90,17 @@ def compute_stats_per_admin(
     for date in ds.time.values:
         date_dt = pd.to_datetime(date)
         if interpolate:
-            output_filename = f"{parameters['iso3_code'].lower()}"
-            f"_seasonal-monthly-single-levels_v5_interp_{date_dt.year}"
-            f"_{date_dt.month}_adm{adm_level}_stats.csv"
+            output_filename = (
+                f"{parameters['iso3_code'].lower()}"
+                f"_seasonal-monthly-single-levels_v5_interp_{date_dt.year}"
+                f"_{date_dt.month}_adm{adm_level}_stats.csv"
+            )
         else:
-            output_filename = f"{parameters['iso3_code'].lower()}"
-            f"_seasonal-monthly-single-levels_v5_{date_dt.year}"
-            f"_{date_dt.month}_adm{adm_level}_stats.csv"
+            output_filename = (
+                f"{parameters['iso3_code'].lower()}"
+                f"_seasonal-monthly-single-levels_v5_{date_dt.year}"
+                f"_{date_dt.month}_adm{adm_level}_stats.csv"
+            )
         output_path = os.path.join(
             country_data_processed_dir, "ecmwf", output_filename
         )
@@ -107,7 +111,6 @@ def compute_stats_per_admin(
                 " skipping"
             )
         else:
-            print(date)
             ds_sel = ds.sel(time=date)
             df = compute_zonal_stats(
                 ds_sel,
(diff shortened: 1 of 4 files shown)


src/utils_general/statistics.py
```diff
@@ -109,3 +111,38 @@ def calc_mpe(observations: np.array, forecast: np.array) -> float:
         / len(observations.time)
         * 100
     )
+
+
+def calc_crps(
+    observations: xr.DataArray,
+    forecasts: xr.DataArray,
```
Tinkaa, line 118: this is the only place where xarray is used. @turnerm Do you know what is the best practice when only using a package to define the type?

turnerm: I think it's totally legit!

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -262,6 +263,54 @@ def convert_dict_to_da(da_dict):
...
+    da_observations: xr.DataArray,
+    da_forecasts: xr.DataArray,
+    normalization: str = None,
+    thresh: float = None,
+) -> pd.DataFrame:
+    """
+    Assumes there is no missing data in da_observations or da_forecasts
+    :param da_observations: data-array or data-set with observed values
+    :param da_forecasts: data-array with forecasted values
+    normalization: (optional) Can be 'mean' or 'std', reanalysis metric
```
turnerm, line 276: this should start with `:param` (otherwise PyCharm complains)

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -262,6 +263,54 @@ def convert_dict_to_da(da_dict):
...
+def get_crps_ecmwf(
+    da_observations: xr.DataArray,
+    da_forecasts: xr.DataArray,
+    normalization: str = None,
+    thresh: float = None,
+) -> pd.DataFrame:
+    """
+    Assumes there is no missing data in da_observations or da_forecasts
+    :param da_observations: data-array or data-set with observed values
+    :param da_forecasts: data-array with forecasted values
```
turnerm, line 275: Maybe specify that obs and forecasts should both be the same length

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -262,6 +263,54 @@ def convert_dict_to_da(da_dict):
...
+    normalization: (optional) Can be 'mean' or 'std', reanalysis metric
+    to divide the CRPS
+    param: threshold: (optional) only select values smaller or equal to
+    this number
+    :return: DataFrame with leadtime index containing the crps
+    """
+    leadtimes = da_forecasts.leadtime.values
+    df_crps = pd.DataFrame(index=leadtimes)
+    observations = da_observations.dropna(dim="time", how="all")
```
turnerm, line 284: In the docstring you say no missing data, then why do you drop NA here?

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -262,6 +263,54 @@ def convert_dict_to_da(da_dict):
...
+        forecasts = da_forecasts.sel(leadtime=leadtime).dropna(
+            dim="time", how="all"
+        )
+
+        if thresh is not None:
+            # cannot index on multidimensional arrays,
+            # e.g. when having lon and lat
+            # xr.where does work on multidimensional arrays
+            observations = observations.where(observations <= thresh)
+            forecasts = forecasts.where(observations <= thresh)
```
turnerm, line 296: Don't you need to do this after the next step?

src/indicators/flooding/glofas/utils.py
```diff
@@ -265,13 +259,12 @@ def get_crps(
...
-            crps = (
-                xs.crps_ensemble(
-                    observations, forecast, member_dim="number"
-                ).values
-                / norm
+
+            df_crps.loc[leadtime, station] = calc_crps(
+                observations,
+                forecast,
+                normalization=normalization,
```
turnerm, line 266: So I guess in this case, the normalization would be calculated after any threshold cut and not on the full sample.

What if you also allowed the `normalization` parameter to be a float, that way you could pass a value directly to use (in addition to the current options), and for the case of GloFAS, just to keep the code working the way it was for now, this value would be calculated in the GloFAS method. Does that make sense?

src/utils_general/statistics.py
```diff
@@ -109,3 +111,38 @@ def calc_mpe(observations: np.array, forecast: np.array) -> float:
...
+    normalization: str = None,
+    member_dim: str = "number",
+) -> float:
+    """
+    observations and forecasts must have the same shape in all
+    dimensions except the member_dim dimension
```
turnerm, line 124: And maybe mention that only the forecast should have the ensemble member dimension

turnerm (approved):

src/indicators/drought/ecmwf_seasonal/processing.py
Tinkaa: Hehe oops I was experimenting and then cheating

src/indicators/flooding/glofas/utils.py
Tinkaa: @turnerm you are so creative!

Tinkaa:
@turnerm Thank you! Could you quickly check if the addition of a float as normalization makes sense? If so, I will merge

src/indicators/flooding/glofas/utils.py
```diff
@@ -250,13 +251,14 @@ def get_crps(
             observations = ds_reanalysis[station].reindex(
                 {"time": forecast.time}
             )
+
             if normalization == "mean":
                 norm = observations.mean().values
             elif normalization == "std":
                 norm = observations.std().values
             elif normalization is None:
```
turnerm, line 259: You can take out this last else if because if it's not met, then `None` will be passed to the crps method

src/indicators/flooding/glofas/utils.py
```diff
@@ -250,13 +251,14 @@ def get_crps(
...
                 norm = 1
```
turnerm, line 260: maybe just use `else` instead of checking for `None`, and then set `norm = normalization` (that way other values can be passed through)

turnerm:
> Could you quickly check if the addition of a float as normalization makes sense? If so, I will merge
One small comment but otherwise looks good!

src/indicators/flooding/glofas/utils.py
turnerm: Not sure if that's a good thing here, aha

PR #130, opened 2021-07-26
Feature/npl events by municipality

turnerm:
The Nepal RC sent a file with flood event impact only in the municipalities of interest, this script was an attempt to correlate these events with GloFAS river discharge.


analyses/npl/09_flood_events_by_municipality.md
```diff
@@ -0,0 +1,300 @@
+We have a file from the RCO that has flood eents only in the municipalities
```
hannahker, line 1: Just a lil typo here: 'eents' :)

analyses/npl/09_flood_events_by_municipality.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+Read in events and clean
+
+```python
+df_events = (pd.read_excel(PAST_EVENTS_FILENAME, index_col='S.No.')
+             .rename(columns={'Mun_ETHOS_Code': 'pcode'})
+            )
+# Drop events with no date and convert date column
+df_events = df_events.loc[df_events['Year'] > 0]
```
hannahker, line 79: How many events is this dropping? Might be good to keep track of that as a caveat in any derived results.

analyses/npl/09_flood_events_by_municipality.md
```diff
@@ -0,0 +1,300 @@
...
+Plot event municipalities over the areas of interest
+
+```python
+fig, ax = plt.subplots()
+karnali.plot(ax=ax)
+koshi.plot(ax=ax)
+df_events.boundary.plot(ax=ax, color='r', lw=0.5)
+```
+
+Might be easier to just do a latitude cut. Latitude > 86 is Koshi, < 83 is Karnali.
```
hannahker, line 121: Or a spatial join/intersect? Although this works well enough.

analyses/npl/09_flood_events_by_municipality.md
```diff
@@ -0,0 +1,300 @@
...
+                         'Private House Fully Damaged']
+plot_results = False
+
+df_events_high_impact = pd.DataFrame(columns=['date', 'basin', 'impact_parameter'])
+
+for basin, station in STATIONS.items():
+    # Reindex the events to the reanalysis time
+    # and get the rolling sum of impact
+    df = df_events_final.loc[basin]
+    df = (df.reindex(ds_glofas_reanalysis.time.data, fill_value=0)
```
hannahker, line 156: Am not totally following here on the logic behind the rolling sum. If each row in ```df_events_final``` is a discrete flood event with non consecutive dates, then what's the axis that we're rolling across (so to speak...)?

analyses/npl/09_flood_events_by_municipality.md
```diff
@@ -0,0 +1,300 @@
...
+```python
+# Define GloFAS events
+rp = 2
+ndays = 1 # Number of consecutive days above RP
+days_before_buffer = 5 # Number of days before GloFAS event the flooding event can occur
+days_after_buffer = 60
+
+df_station_stats = pd.DataFrame(columns=['station', 'impact_parameter', 'TP', 'FP', 'FN'])
+
+for basin, station in STATIONS.items():
```
hannahker, line 187: Looks like this is repeated a bunch across notebooks so I think definitely good to generalize.

hannahker (approved):

analyses/npl/09_flood_events_by_municipality.md
turnerm: Ah actually it's not dropping any, this line was a remnant from the old events file. I've removed it, good catch!

analyses/npl/09_flood_events_by_municipality.md
turnerm: I chose the latitude cut because a couple of municipalities in the easternmost group don't quite intersect with the Saptakoshi watershed.

analyses/npl/09_flood_events_by_municipality.md
turnerm: It's just a way of smoothing the data, particularly in the case where there are several small events over the course of a few days, these should be grouped into a larger single event. I hope that makes sense.

analyses/npl/09_flood_events_by_municipality.md
turnerm: Yes totally agree. Now that Nepal is wrapped up I actually have on my to-do list to refactor all the notebooks, in particular:
- Move all the pre-amble stuff to some kind of config file (like all the file locations etc)
- Refactor exactly this method. It's sometimes done slightly differently in some cases but I really want to try and generalize it.

I will make a new PR for this (hopefully today!)

PR #131, opened 2021-07-26
Nepal more detailed trigger analysis

turnerm:
Script to make more detailed historical timeline of past readiness / action triggers. Changes include: 
- Defining the action / readiness trigger as the forecast reaching threshold on days 1-3 / 4-7 (not just single days), which is a more accurate representation of the actual trigger
- Make date shifting optional in the glofas reanalysis loading


analyses/npl/10_historical_event_timeline.md
```diff
@@ -0,0 +1,320 @@
...
+
+fig, ax = plt.subplots()
+x.plot.line(hue='percentile', alpha=0.5)
+ax.set_xlim(np.datetime64('2009-08-10'), np.datetime64('2009-08-20'))
+ax.axhline(rp_val, c='k')
+ax.set_ylim(4000, 8000)
+```
+
+```python
+# Plot timeline of events
```
hannahker, line 280: These figures look really nice! Only suggestion would be to add in a legend to indicate what each of the shaded sections correspond to, but you might also have done this manually in the slides.

hannahker (approved):
Already looked through most of this in #133 so still good to go. 

Not super critical, but I'd be interested to look more into the results for different forecast percentiles to see how that impacts the overall performance.

analyses/npl/10_historical_event_timeline.md
turnerm: There is a legend way at the bottom

turnerm:
> Already looked through most of this in #133 so still good to go.

It's so weird, I didn't actually make any changes to `10_historical_event_timeline.md` in #133 but for some reason they are all showing up in that PR -- very sorry for the confusion! (But thanks for looking at it there)

> Not super critical, but I'd be interested to look more into the results for different forecast percentiles to see how that impacts the overall performance.

Agreed, I think it's a free parameter we haven't explored enough when developing the flood triggers

PR #132, opened 2021-07-26
Feature/mwi ecmwf skill

Tinkaa:
Further explore the skill of ECMWF. 
The only file that is relevant is `07_ecmwf_monthly_skill_general.md`. The others are small changes that don't really need review imo. 
For (0)07, I think it is easier to go through the notebook, than to look at the diff since basically everything changed. 

Main questions:
- Do you think the methods to assess the skill make sense? 
- Do the conclusions make sense? 
- Do you see potential bugs?
- Anything missing? 

Soon another notebook will come which looks at the admin level instead of cell level (and has the scrollable plots😁)


Tinkaa:
@turnerm sorry I meant to make a separate PR but I messed up so hope it is okay like this. 
- `08_ecmwf_monthly_skill_admin1.md` does basically the same as 07 but then at admin1 level instead of cell level. 
- Small changes in `src/indicators/drought/ecmwf_seasonal/processing.py ` and `src/malawi/get_ecmwf_seasonal_data.py` since I noticed I omitted computing the stats per admin. These scripts are not perfect, but want to do a larger refactor at some point so not putting too much effort into it now

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -16,22 +19,15 @@ from pathlib import Path
...
 
-import rioxarray
+import xarray as xr
 import matplotlib.pyplot as plt
 import matplotlib as mpl
 import pandas as pd
-import xskillscore as xs
 import numpy as np
-
+import plotly.express as px
```
turnerm, line 27: add plotly to requirements

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -41,164 +37,372 @@ reload(processing)
...
-# Determine general skill per leadtime
+```python
+da_obs=xr.load_dataset(chirps_monthly_mwi_path)
+da_obs=da_obs.precip
+```
 
+```python
+#interpolate forecast data such that it has the same resolution as the observed values
+#using "nearest" as interpolation method and not "linear" because the forecasts are designed to have sharp edged and not be smoothed
+da_forecast=da_lt.interp(latitude=da_obs["latitude"],longitude=da_obs["longitude"],method="nearest")
```
turnerm, line 79: What if instead of interpolating the forecast, you down-sampled the observations?

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -41,164 +37,372 @@ reload(processing)
...
+
+Next we look at the bias. We firstly plot the observed vs forecasted-observed values across all leadtimes, dates, and cells. 
+From this we can see that
+
+- Most months with very low precipitation were correctly classified. 
+- Months with less than 300mm have the tendency to be overpredicted, i.e. we see a positive bias
+- Months with more than 300mm have the tendency to be underpredicted, i.e. we see a negative bias
+
+```python
+#plot the observed vs forecast-observed to get a feeling for the discrepancy between the two
```
turnerm, line 275: It's also interesting to plot just forecast vs observed. It just looks like a giant blob, which is not good. For Jan Feb you can see that the forecast histogram is shifted to higher values than the observed one.

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -41,164 +37,372 @@ reload(processing)
...
+
+```python
+def calc_diff_stats(observations,forecast):
+    diff=forecast-observations
+    diff_mean=diff.mean()
+    diff_median=diff.median()
+    diff_std=diff.std()
+    diff_count=diff.count()
+    ci95_hi = diff_mean + 1.95*diff_std/math.sqrt(diff_count)
+    ci95_lo = diff_mean - 1.95*diff_std/math.sqrt(diff_count)
```
turnerm, line 336: The confidence interval seems a bit too small, I think because the points in your sample are not independent. The best way to get a confidence interval here would be to use the ensemble distribution

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -41,164 +37,372 @@ reload(processing)
...
+ax.plot(df_diff.index,df_diff["all_mean"],label="mean")
+ax.fill_between(df_diff.index, df_diff.all_ci_lo, df_diff.all_ci_hi,alpha=0.2,label="confidence interval")
+ax.plot(df_diff.index, df_diff["all_med"],label="median")
+ax.set_xlabel("leadtime [months]")
+ax.set_ylabel("median bias")
+ax.grid()
+ax.legend(bbox_to_anchor=(1.05, 1));
+```
+
+When looking per month though, we can see that the average bias clearly differs per month. The months around the rainy season have the highest bias (nov, dec, jan, feb). This can either be because values are generally larger, or because the forecast has less skill.
```
turnerm, line 381: I'm not sure I understand exactly what's plotted here. When I took a quick peak at a subset of the data (e.g. `df_forobs[(df_forobs['month']==1) & (df_forobs['leadtime']==1)]['diff_forobs']`) the mean / median are closer to 40 / 50.

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -41,164 +37,372 @@ reload(processing)
...
 
+From the plots below we can see that
+- The CRPS is pretty high, resulting in a 30% error across all values
+- The CRPS is relatively steady across leadtimes. We can see a dip at 1 month leadtime, and a smaller dip at 4 months leadtime
+- The CRPS differs across months and thresholds
+- For the months around the rainy season (Nov-Apr), we can see that the CRPS is relatively high, but the normalized CRPS is lower. This is largely due to the fact that the precipitation during the rainy season is higher. 
+    - While lower during Nov-Apr, and specifically JanFeb, the error is still 25% which is quite large. 
+- When selecting on the precipitation threshold, we can see that the higher the threshold the lower the normalized CRPS. 
+    - This is likely because there are many very low values in the data, which causes the normalized CRPS to get large quickly even though the absolute difference might be small
+
```
turnerm, line 163: It's weird that the CRPS doesn't get worse with lead time. Maybe that's a sign that it's not an appropriate metric here

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -41,164 +37,372 @@ reload(processing)
...
-# reduced skill (dividing by standard devation).
-plot_skill(df_crps, division_key='std', ylabel="RCRPS")
-
-#This is perhpas not exactly what we want because we know this 
-#data comes from the same location and the dataset has the same properties, 
-#but we are splitting it up by mean value. Therefore try normalizing using mean
-plot_skill(df_crps, division_key='mean', ylabel="NCRPS (CRPS / mean)")
+Below we plot the range of observed and forecasted values per month, but across all leadtimes. We can see that they both show the same pattern across the months. However, the observed data shows a broader range of values, and especially has more outliers. The medians are relatively in the same range for observed and forecasted values, though when zooming in you can see that they also differ. 
+
+```python
```
turnerm, line 251: It might be useful to split up this plot (and the ones below) by leadtime, just to see if there is any difference. It should be more difficult to forecast 6 months ahead than 1 month ahead, right?!

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
```diff
@@ -41,164 +37,372 @@ reload(processing)
...
+px.line(df_diff.reset_index(), 
+        x='leadtime', 
+        y=[c for c in df_diff.columns if "thresh" in c and "med" in c]+["all_med"],
+        labels={"value": "median bias"}
+       )
+```
+
+Questions:
+- The values we saw in the CRPS plots are a lot larger than for the median bias plots. Why is this the case?
```
turnerm, line 402: The bias can be lower because it sums both negative and positive errors, which can cancel each other out. With the CRPS the errors are squared.

turnerm (approved):

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
Tinkaa: Hmm nice idea. The only thing being that down-sampling the observations results in quite some NaN's due to the way the observational data is clipped.. What would you say is the advantage of down-sampling instead of up-sampling?

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
Tinkaa: image
 I agree it is a giant blob, but don't think it gets better when plotting for vs obs? (see image). Would it be an idea to remove all observed values below 20 mm  as that is the majority of points while we are not really interested in them? Or is that cheating

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
Tinkaa: Good idea! My reasoning for down-sampling instead of upsampling would be that the forecast can only predict at the resolution it has, so if comparing with a higher resolution it is unfair cause it will always be off for some parts then. Would you agree on that?

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
Tinkaa: Smart on the not independence! 
Regarding ensemble distribution, I am not sure. As we want to aggregate them to one number and here we want to know how well that aggregation works. Why would you say to use the ensemble distribution?

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
Tinkaa: That is super strange. Cause that is exactly what the plot shows, but when I do `df_forobs[(df_forobs['month']==1) & (df_forobs['leadtime']==1)]['diff_forobs'].median()` , I do get the values as shown in the plot (in this case 26.64). Could you try again?

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
Tinkaa: I agree it is really weird. At the same time, also when looking at other metrics it doesn't get worse. E.g. for the trigger design the performance of 3 months leadtime was better than that of 2 months. I think this is something we should ask ECMWF.

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
Tinkaa: Haha yes it should be. 
What I struggle with is that it becomes so much information, which is hard to compare. That is why I tried to summarize the differences in leadtime by the bias. 
I tried separating these plots by leadtime. You do see some differences especially 1 months vs all other months. But comparison is really hard

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
turnerm: Yes agreed, it will give you a better picture of what to expect for the trigger since ultimately you will be using the forecast at the given resolution.

Also, when you interpolate it isn't really the same as increasing the resolution; there are effects to keep in mind, like you will be more likely be missing extreme values in the interpolated dataset.

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
turnerm: Is what's plotted there really forecast vs obs? Anyway I was just curious to make that plot to see if there is any obvious relation between the two, or if it's more just random looking, which means the forecast isn't forecasting much of anything. 

> Would it be an idea to remove all observed values below 20 mm as that is the majority of points while we are not really interested in them?

I thought we are interested in these?

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
turnerm: Hm, as I'm still a bit confused about what's plotted here maybe we could talk about it

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
turnerm: (Oh and it will also make it way less memory intensive to run!)

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
turnerm: I plotted again and it looks much more reasonable: 
image

analyses/malawi/dryspells_trigger/07_ecmwf_monthly_skill_general.md
turnerm: For sure it's a lot of information. But I think it's important to keep the variables separate if you can, to avoid ambiguous / difficult to interpret results due to competing effects between variables. It might help to focus just on one or two lead times that you are interested in, for example

Tinkaa:
@turnerm I added the newest analyses, with the graphs now included in the slides and some other things. Do you think it is good to go to merge?

PR #133, opened 2021-07-27
Nepal forecast return period

turnerm:
Script to calculate the equivalent return period in the forecasts

### DIFF
--- a/analyses/npl/11_forecast_rp.md
+++ b/analyses/npl/11_forecast_rp.md
@@ -97,5 +97,5 @@ for station in STATIONS:
     rp_val = df_return_period.loc[MAIN_RP, station[:-3]]
     for leadtime, df_rps_forecast in df_rps_forecast_dict.items():
         rp_equiv = df_rps_forecast.index[(df_rps_forecast[station] - rp_val).abs().argmin()]
-        print(leadtime, rp_equiv)
+        print(leadtime, f'{rp_equiv:.1f}')
 ```
--- a/analyses/npl/11_forecast_rp.md
+++ b/analyses/npl/11_forecast_rp.md
@@ -1,3 +1,7 @@
+Because the forecasts at Chatara and Chisapani have a negative bias, 
+we want to know what to know what a 2 year RP effectively corresponds
+to for each lead time.
+
 ```python
 from pathlib import Path
 import os
--- a/analyses/npl/11_forecast_rp.md
+++ b/analyses/npl/11_forecast_rp.md
@@ -4,32 +4,93 @@ import os
 import sys
 from importlib import reload
 
+import numpy as np
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+import pandas as pd
+
 path_mod = f"{Path(os.path.dirname(os.path.realpath(''))).parents[0]}/"
 sys.path.append(path_mod)
 
 from src.indicators.flooding.glofas import utils
 reload(utils)
+mpl.rcParams['figure.dpi'] = 200
 ```
 
 ```python
 COUNTRY_ISO3 = 'npl'
 LEADTIMES = [x+1 for x in range(7)]
+STATIONS = [
+    'Chatara_v3',
+    'Chisapani_v3'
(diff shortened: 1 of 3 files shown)


analyses/npl/10_historical_event_timeline.md
```diff
@@ -0,0 +1,307 @@
...
+        ax.grid()
+        # Add the activations
+        for event_type, hatch, colour, lw in zip(['readiness', 'action','danger'], 
+                                             ['/////', '\\\\\\\\\\', '.....'], 
+                                             ['C2', 'C1', 'C6'],
+                                             [0.5, 0.5, 2]):
+            cname = f'event_{event_type}'
+            if event_type != "danger":
+                cname += f"_rp{MAIN_RP}"
+            ax.fill_between(data.index, y0, y1, where=data[cname],
```
hannahker, line 297: Am just getting ```KeyError: 'event_readiness_rp2'``` when trying to rerun this notebook.

analyses/npl/10_historical_event_timeline.md
```diff
@@ -0,0 +1,307 @@
...
+# Find TP, FP and FN
+
+def get_consecutive_groups(x: pd.Series, n=2):
+    # Find indices whre False / True change. Skip the first entry of the array because it always changes there
+    l = np.where(x.shift() != x)[0][1:]
+    # Convert array to pairs
+    return [l[i:i+n] for i in range(0, len(l), n)]
+
+
+def get_station_stats(df_station_dict, event_var, rp=MAIN_RP):
```
hannahker, line 149: I think this would definitely be good to generalize and add to the general glofas ```utils.py``` (probably improving on what I've done in #135).

analyses/npl/10_historical_event_timeline.md
```diff
@@ -0,0 +1,307 @@
...
+                    max_leadtime = -1
+                    for leadtime in leadtimes:
+                        if df_station.loc[glofas_event, f'forecast_lt{leadtime}_p{percentile}'] >= rp_val:
+                            max_leadtime = leadtime
+                    if max_leadtime == -1:
+                        print('Something went wrong!')
+                    # Check if any events are around that date
+                    days_offset = (df_true_events.index - glofas_event) /  np.timedelta64(1, 'D')
+                    # Add the max leadtime to the days after buffer
+                    detected = (days_offset >= -1 * days_before_buffer) & (days_offset <= days_after_buffer + max_leadtime)
```
hannahker, line 179: Also an interesting way to look at detections that is slightly different than the approach that I've been taking in my work. This looks good to me but it would be great to have a quick chat on it together so that we're generally taking the same approach :)

hannahker (approved):
Looks good to me. Main thing I'd like to follow up on with you is generalizing/standardizing an approach for calculating the detection statistics from GloFAS events. Also I think generally it would be good practice to keep 1-2 sentences at the top of each notebook to just give a quick overview of what the main intent of the notebook is.

analyses/npl/10_historical_event_timeline.md
turnerm: Ah sorry, this has been fixed in #131 which was supposed to be the main branch for this file. Not sure why it's showing up here!

analyses/npl/10_historical_event_timeline.md
turnerm: Agreed 110%, new PR coming up ASAP!

analyses/npl/10_historical_event_timeline.md
turnerm: For sure, I also realized that there are a couple of ways that you could define TPs depending on what you want. I will ping you and we can plan a chat

turnerm:
> Also I think generally it would be good practice to keep 1-2 sentences at the top of each notebook to just give a quick overview of what the main intent of the notebook is.

Definitely, you are always very good at documenting your notebooks and it's definitely something I need to improve upon! I will add some more documentation to the RP notebook

PR #134, opened 2021-07-29
Proof of concept: top-level entry script

turnerm:
It's a bit rough around the edges, but here is what a top-level entry script just for GloFAS could look like. 

Some things to address:

- it's now in the very top level of the repo, which could get messy if we have many of these
- I didn't find an elegant way to pass (usually unused) kwargs to the GloFAS methods (here it could be good to have defaults for the config files, that will take some setting up)


turnerm:
Closing because switching to [aa-toolbox](https://github.com/OCHA-DAP/pa-aa-toolbox)

PR #144, opened 2021-08-04
Nepal refactor

turnerm:
- Created a `npl_settings` file that has common paths and a couple of settings that are used in most notebooks
- Refactored notebooks to use the settings file
- Created / modified a couple of the stat calculation methods (for TP, FP etc) in GloFAS utils
- Invoked the use of this method in the relevant notebooks
- Added description to the top of each notebook


analyses/npl/02_glofas_skill.md
```diff
@@ -1,50 +1,30 @@
...
-    'Karnali': ['Chisapani_v3', 'Asaraghat_v3', 'Dipayal_v3', 'Samajhighat_v3'],
-    'Rapti': ['Kusum_v3'],
-    'Bagmati': ['Rai_goan_v3'],
-    'Babai': ['Chepang_v3']
-}
-STATIONS_BY_MAJOR_BASIN = {
-    'Koshi': ['Chatara_v3', 'Simle_v3', 'Majhitar_v3', 'Kampughat_v3', 'Rai_goan_v3'],
-    'Karnali': ['Chisapani_v3', 'Asaraghat_v3', 'Dipayal_v3', 'Samajhighat_v3', 'Kusum_v3', 'Chepang_v3'],
-}
+MAIN_RP = 1.5 # RP used as a threshold to get extreme values
```
hannahker, line 20: Would it make sense to put this in the ```settings.py``` file too, or do you think this would change too much between notebooks?

analyses/npl/04_glofas_flood_events.md
```diff
@@ -173,12 +119,13 @@ for istation, station in enumerate(FINAL_STATIONS):
 ```python
 # Print out some stats
 # Round to the nearest 5 for a presentation (10 seems a bit too coarse)
+rp = 1.5
 def round_to_5(x):
```
hannahker, line 123: Would this make sense to instead put as an option in ```utils.get_more_detection_stats()```?

analyses/npl/09_water_level_events.md
```diff
@@ -207,26 +165,17 @@ for station in STATIONS:
             print(f"{station}")
             print(f"True events: {len(df_true_events)}")
```
hannahker, line 166: Think this should be changed to ```wl_events```? 

Tried running this notebook and got ```NameError: name 'df_true_events' is not defined```

analyses/npl/npl_settings.py
```diff
@@ -0,0 +1,86 @@
+import os
```
hannahker, line 1: So great to have this file and keep things cleaner in the notebooks! For my own curiosity, what's the advantage of keeping this as ```.py``` rather than something like a ```config.yml```?

Also I'd vote to refer to these are parameters rather than settings, but not really a big deal.

src/indicators/flooding/glofas/utils.py
```diff
@@ -303,87 +307,84 @@ def get_groups_above_threshold(
...
-    df_glofas_act = pd.DataFrame(groups, columns=["start_index", "end_index"])
-    df_glofas_act["num_days"] = (
-        df_glofas_act["end_index"] - df_glofas_act["start_index"]
-    )
-    df_glofas_act["start_date"] = df_glofas_act["start_index"].apply(
-        lambda x: da_glofas.time[x].values
-    )
-    df_glofas_act["end_date"] = df_glofas_act["end_index"].apply(
-        lambda x: da_glofas.time[x].values
+def get_dates_list_from_dataset(
```
hannahker, line 310: Since we talked about it today, could we add a docstring to this function? :)

src/indicators/flooding/glofas/utils.py
```diff
@@ -303,87 +307,84 @@ def get_groups_above_threshold(
...
+        threshold=threshold,
+        min_duration=min_duration,
     )
-    return df_glofas_act
+    return [da.time[group[0] + min_duration - 1].data for group in groups]
 
 
-def get_detection_stats(df_glofas, df_impact, buffer_before, buffer_after):
-    TP = 0
+def get_detection_stats(
```
hannahker, line 321: Great to have this here, and implemented so much simpler than I had done before!

hannahker (approved):
Very nice :) 

I didn't rerun all of the notebooks, but just picked out a couple of things that I noticed. I think you also missed adding some intro documentation to ```08_water_level.md```. Great to have this refactoring overall and I really like the addition of the settings file. 

Thanks also for generalizing those detection stats functions and improving on what I'd done before!

analyses/npl/02_glofas_skill.md
turnerm: I have `MAIN_RP = 2` in the settings file, but in this analysis I want to use 1.5 because the RP Is just used as a cutoff for "high" values, so that's why it's included explicitly here. Maybe there is a better way to make it clear that this is an override?

analyses/npl/04_glofas_flood_events.md
turnerm: I put the rounding function in general utils, but I kept it out of `get_more_detection_stats` because I think you may want to keep the parameters there in fraction form for plotting (and can always round by hand later)

analyses/npl/09_water_level_events.md
turnerm: Ah oops you're right, sorry about that! I think I was running this without restarting the kernel and it was using a cached `df_true_events`.

analyses/npl/npl_settings.py
turnerm: >  For my own curiosity, what's the advantage of keeping this as .py rather than something like a config.yml?

I guess the advantages of a .py file are:
- easier to import (no mucking around with yaml)
- more flexible format, you can add methods or imports as I did
I prefer to use a yaml file when the hierarchical structure becomes more useful, such as when you have multiple parameter files for different countries but with the same parameter names, for example. Hope that makes sense1

> Also I'd vote to refer to these are parameters rather than settings, but not really a big deal.

Good call, changed the name!

src/indicators/flooding/glofas/utils.py
turnerm: Good call, done!

Issue #94, opened 2021-06-08
Utils function for computing stats within polygon


Tinkaa:
For many datasets we are computing statistics of raster files within a region. 
We have an outdated function for this in `general_utils` named `compute_raster_statistics`. But in practice we (or I) have been defining a new function for each use case. 

So goal would be to see if we can capture all use cases in one general function. 

Also links to #85 to try to make computing these stats faster.

turnerm:
@Tinkaa can this be closed now?

Tinkaa:
Yes addressed in #146

Issue #85, opened 2021-05-27
[ecmwf] test rasterio.mask instead of zonal_stats


Tinkaa:
[The processing script](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/f8c6400ab79531560c25e8a371313da7719110a7/src/indicators/drought/ecmwf_seasonal/processing.py) is currently using rasterstats.zonal_stats to compute stats per admin area. This is rather slow. So we should investigate if there is a faster method. 
Rasterio.mask might be an option, as indicated [here](https://github.com/OCHA-DAP/pa-COVID-model-parameterization/issues/35). Though this requires the rasterdata to be loaded as rasterio object. [This SO](https://gis.stackexchange.com/questions/357490/mask-xarray-dataset-using-a-shapefile) also has some suggestions

Tinkaa:
For future reference: rasterstats and clipping are equally slow after all, it just depends whether the DataArray has already been loaded into memory. Which is done when one of the two methods is called, so it just depends on the order which one seems faster. I also experimented with loading the DataArray first or let the function to do that and there is no time difference. So it is up to the user to load the DataArray on beforehand or not, depending on further computations.

PR #146, opened 2021-08-06
example stats rioxarray

Tinkaa:
We regularly need stats per admin area retrieved from a raster file. 
So far we have mainly been using the `zonal_stats` function from `rasterstats`. While it is a great function, it is very slow. 
Another method is to clip the raster to the area of interest with `rio.clip` and then compute the stats
From quick testing this gives the same results and is waaay faster. 

My question is if you think this is a valid method or anything should be improved? 
This is just a notebook showing both methods. It is not meant to be merged into `develop` and is messy. But wanted to do a quick feedback loop, and if you think this works, I I will implement it as a proper function in `general_utils`

This links to issues #85 and #94


turnerm:
This looks amazing, it's sooo much faster. I don't see any issues, but there are a couple of things that I would want to understand (and I'm being lazy and just asking you):
- Do you know why the rioxarray method is so much faster than zonal stats?
- Do these methods preserve the totals? (e.g. if you took the sum of each region, would that then add up to the full raster?)
- Related: How do they deal with edges? 

For the implementation in `general_utils` I think this would be a really good method to add some tests too, just because it will be so widely used and we will really want to be sure that it's working correctly.

turnerm (approved):

Tinkaa:
@turnerm 
> Do you know why the rioxarray method is so much faster than zonal stats?

I have no clue and that is a good question. I have browsed a bit through their methods and see if there has been a comparison on SO, but couldn't find any.. I think zonal stats has a bit more options but still. 

> How do they deal with edges?
You can choose to include all cells with their center within the edges, or all cells touching with any part in within the edges, idenitified by the `all_touched` argument. See a bit more explanation [here](https://pythonhosted.org/rasterstats/manual.html#rasterization-strategy)

> Do these methods preserve the totals? (e.g. if you took the sum of each region, would that then add up to the full raster?)

Good question and yes they do! Or well depending how edges are handled, the totals are preserved if only taking into account cells with their centre inside the region. If taking all cells that touch the region then no as when splitting up the raster into regions in this case cells might be counted double

Good idea on tests, now I don't have excuses anymore to learn that

Tinkaa:
[Question I asked on SO on this](https://stackoverflow.com/questions/68726522/compute-polygon-mean-from-raster-rasterstats-zonal-stats-versus-rio-clip)

Tinkaa:
@turnerm would you be willing to review `src/utils_general/raster_manipulation.py`? 
Switch back to using rasterstats, assuming they will continue to optimize that. Added the fact that there is no time difference between the two methods to #85

Tinkaa:
@turnerm it is ready to be reviewed!
I am using the clip method now instead of zonal_stats, because we still don't understand the zonal_stats bug.
However, with this method the function becomes a lot longer and it might not handle edge cases. For example if an admin area is completely outside the raster. 
So I am still in doubting mode what is better. The advantage of clip is definitely that it seems to be more careful with strange coordinate system things.

src/utils_general/raster_manipulation.py
```diff
@@ -1,8 +1,9 @@
 import logging
 import geopandas as gpd
-import numpy as np
-from rasterstats import zonal_stats
 import pandas as pd
+import xarray as xr
+from typing import List
+import rioxarray  # noqa: F401
```
turnerm, line 6: Imports should be grouped into standard and external libraries, see [here](https://www.python.org/dev/peps/pep-0008/#imports)

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
-        )
-    )
-    df["perc_threshold_touched"] = (
-        bin_zonal_touched["sum"] / bin_zonal_touched["count"] * 100
-    )
-
-    return df
+    df_list = []
+
+    for a in gdf[bound_col].unique():
```
turnerm, line 139: Could you use a more descriptive variable name for `a`, like `row`, or `admin_region`?

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+        gdf_adm = gdf[gdf[bound_col] == a]
+
+        da_clip = raster_array.rio.set_spatial_dims(
+            x_dim=lon_coord, y_dim=lat_coord
+        ).rio.clip(gdf_adm[geom_col], all_touched=all_touched)
+        if stats_list is None:
+            stats_list = ["mean", "std", "min", "max", "sum", "count"]
+
+        grid_stat_all = []
+        for s in stats_list:
```
turnerm, line 149: Better to call it `stat` (or something more descriptive), rather than `s`

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+                grid_stat = getattr(da_clip, s)(dim=[lon_coord, lat_coord])
+            else:
+                # if array only contains NaNs, "sum" will return 0
+                # while NaN might be preferred
+                # this can be fixed by setting the min_count arg
+                # but this arg is not present in the other function calls.
+                # therefore chose to leave it as it is
+                grid_stat = getattr(da_clip, s)(
+                    dim=[lon_coord, lat_coord], skipna=True
+                )
```
turnerm, line 163: To make this more succinct, you could do something like:
```python
kwargs = {} if s == 'count' else {'skipna': True}
grid_stat = getattr(da_clip, s)(dim=[lon_coord, lat_coord], **kwargs)
```

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+        for s in stats_list:
+            if s == "count":
+                # count automatically ignores NaNs
+                # therefore skipna can also not be given as an argument
+                # implemented count cause needed for computing percentages
+                grid_stat = getattr(da_clip, s)(dim=[lon_coord, lat_coord])
+            else:
+                # if array only contains NaNs, "sum" will return 0
+                # while NaN might be preferred
+                # this can be fixed by setting the min_count arg
```
turnerm, line 158: If you have `kwargs` as a separate variable, you could add `min_count` as a key only if `s == "sum"`

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
-    return df
+    df_list = []
+
+    for a in gdf[bound_col].unique():
+        gdf_adm = gdf[gdf[bound_col] == a]
+
+        da_clip = raster_array.rio.set_spatial_dims(
+            x_dim=lon_coord, y_dim=lat_coord
+        ).rio.clip(gdf_adm[geom_col], all_touched=all_touched)
+        if stats_list is None:
```
turnerm, line 145: This should be initialized at the beginning of the method (and outside of the for loop) since it's setting the default

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+                # if array only contains NaNs, "sum" will return 0
+                # while NaN might be preferred
+                # this can be fixed by setting the min_count arg
+                # but this arg is not present in the other function calls.
+                # therefore chose to leave it as it is
+                grid_stat = getattr(da_clip, s)(
+                    dim=[lon_coord, lat_coord], skipna=True
+                )
+            grid_stat = grid_stat.rename(f"{s}_{bound_col}")
+            grid_stat_all.append(grid_stat)
```
turnerm, line 165: Instead of appending to a list, it might make sense to directly create a dictionary, which will automatically keep track of the variable name. If you set `grid_stat_all = {}`, then:
```python
grid_stat[s] = getattr(da_clip, s)(dim=[lon_coord, lat_coord], **kwargs).values
```

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+                grid_stat_all.append(grid_quant)
+        # if dims is 0, it throws an error when merging
+        # and then converting to a df
+        # this occurs when the input da is 2D
+        if grid_stat_all[0].dims == ():
+            df_adm = pd.DataFrame(
+                {da_stat.name: [da_stat.values] for da_stat in grid_stat_all}
+            )
+        else:
+            zonal_stats_xr = xr.merge(grid_stat_all)
```
turnerm, line 181: I was wondering, in which cases would the dimensions not be equal to `()`?

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+                grid_quant = grid_quant.drop("quantile").rename(
+                    f"{q}quant_{bound_col}"
+                )
+                grid_stat_all.append(grid_quant)
+        # if dims is 0, it throws an error when merging
+        # and then converting to a df
+        # this occurs when the input da is 2D
+        if grid_stat_all[0].dims == ():
+            df_adm = pd.DataFrame(
+                {da_stat.name: [da_stat.values] for da_stat in grid_stat_all}
```
turnerm, line 178: Will `values` always be a length 1 array? Should we always take the 0th element? I guess this is anyway simplified by pandas

src/utils_general/raster_manipulation.py
turnerm: Actually, since you are converting to a dataframe later, you could just use one from the start, not sure what would be cleaner

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+                # this can be fixed by setting the min_count arg
+                # but this arg is not present in the other function calls.
+                # therefore chose to leave it as it is
+                grid_stat = getattr(da_clip, s)(
+                    dim=[lon_coord, lat_coord], skipna=True
+                )
+            grid_stat = grid_stat.rename(f"{s}_{bound_col}")
+            grid_stat_all.append(grid_stat)
+        if percentile_list is not None:
+            for q in percentile_list:
```
turnerm, line 167: This could be turned into a list comprehension

src/utils_general/raster_manipulation.py
turnerm: Or why not iterate over the dataframe rows?

tests/test_raster_manipulation.py
```diff
@@ -0,0 +1,48 @@
...
+        "name": ["hi", "bye"],
+        "geometry": [
+            Polygon([(0, 0), (0, 2), (2, 2), (2, 0)]),
+            Polygon([(2, 0), (2, 2), (3, 2), (3, 0)]),
+        ],
+    }
+    gdf = gpd.GeoDataFrame(d)
+
+    df_stats = compute_raster_statistics(
+        gdf, "name", da, stats_list=["min", "max", "mean", "count", "sum"]
```
turnerm, line 28: Could you also throw in a couple of percentiles to test ;)

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +100,88 @@ def fix_calendar(ds, timevar="F"):
...
+                grid_quant = da_clip.quantile(q, dim=[lon_coord, lat_coord])
+                grid_quant = grid_quant.drop("quantile").rename(
+                    f"{q}quant_{bound_col}"
```
turnerm, line 170: It might be tricky to work with a decimal in the column name. Would it make more sense for the user to pass in integer percentiles?

src/utils_general/raster_manipulation.py
turnerm: Final thought on this: do you think that it would make sense to add the stats as new columns in the gdf data frame itself?

src/utils_general/raster_manipulation.py
Tinkaa: Wooowww this is amazing, thanks!

src/utils_general/raster_manipulation.py
Tinkaa: If you have a 4D dataframe. E.g. lon,lat,precip and time. Then you can directly loop over all the times. Does that make sense?

src/utils_general/raster_manipulation.py
Tinkaa: Hmm I don't think so because of the thing discussed above if the data is 3D. In that case I am merging a list and not converting to a df directly. Does that make sense?

src/utils_general/raster_manipulation.py
Tinkaa: yes due to this if check, the values should always be 1. 
Do you mean that we should check if the length is 1?

src/utils_general/raster_manipulation.py
Tinkaa: How?

tests/test_raster_manipulation.py
Tinkaa: Good one, done. With this small number of data points the quantiles are a bit funny, as some linear interpolation is being done. But I do think that is the most valide method (you could also e.g. take the nearest or max value without interpolation)

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +101,90 @@ def fix_calendar(ds, timevar="F"):
...
+        da_clip = raster_array.rio.set_spatial_dims(
+            x_dim=lon_coord, y_dim=lat_coord
+        ).rio.clip(gdf_adm[geom_col], all_touched=all_touched)
+
+        grid_stat_all = []
+        for stat in stats_list:
+            # count automatically ignores NaNs
+            # therefore skipna can also not be given as an argument
+            # implemented count cause needed for computing percentages
+            kwargs = {} if stat == "count" else {"skipna": True}
```
turnerm, line 155: This format is good if there is only one difference from the default, but since you also have a difference for "sum", I would structure them both the same way (i.e. set `kwargs = {}` and then on a separate line check `if stat == "count"` etc)

src/utils_general/raster_manipulation.py
turnerm: Ah okay yes makes sense.

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +101,90 @@ def fix_calendar(ds, timevar="F"):
...
+            # makes sum return NaN instead of 0 if array
+            # only contains NaNs
+            if stat == "sum":
+                kwargs["min_count"] = 1
+            grid_stat = getattr(da_clip, stat)(
+                dim=[lon_coord, lat_coord], **kwargs
+            )
+            grid_stat = grid_stat.rename(f"{stat}_{bound_col}")
```
turnerm, line 163: I would chain this line with the previous one

src/utils_general/raster_manipulation.py
turnerm: Ah okay makes sense. In the future it might be worth writing a simple test to make sure that the 3d case works (but it's fine for now).

src/utils_general/raster_manipulation.py
```diff
@@ -99,111 +101,90 @@ def fix_calendar(ds, timevar="F"):
...
+                )
+                grid_quant = grid_quant.drop("quantile").rename(
+                    f"{quant}quant_{bound_col}"
+                )
+                grid_stat_all.append(grid_quant)
+
+        # if dims is 0, it throws an error when merging
+        # and then converting to a df
+        # this occurs when the input da is 2D
+        if grid_stat_all[0].dims == ():
```
turnerm, line 179: This could be made slightly more general by writing `if not grid_stat_all[0].dims` (that way if dims is a different type but also empty it will still work)

src/utils_general/raster_manipulation.py
turnerm: I don't remember what I had in mind here :sweat_smile:

src/utils_general/raster_manipulation.py
turnerm:
```python
        grid_stat_all = [da_clip.quantile(quant / 200, dim=[lon_coord, lat_coord])
                             .drop("quantile")
                             .rename(f"{quant}quant_{bound_col}") 
                         for quant in percentile_list]
```

turnerm:
Re-approved ;)

PR #149, opened 2021-08-10
Add notebook to visualize glofas around flood onset

hannahker:
Just adding an exploratory notebook to visualize GloFAS values around the time of a flood onset.

### DIFF
--- a/analyses/malawi/flooding_trigger/08_mwi_event_visualization.md
+++ b/analyses/malawi/flooding_trigger/08_mwi_event_visualization.md
@@ -62,7 +62,7 @@ for station in stations_adm2.values():
 Create a plot for each event. For the dates around each flood onset, we want to see what both the GloFAS reanalysis (historical) and reforecast data look like. We'll pick a specific forecast probability and include forecasts from various leadtimes. The y-axis is scaled to the percentage of a given return period threshold. Lines on the graph have a stronger weight when the values exceed the threshold.
 
 ```python
-buffer_start = 5
+buffer_start = 15
 buffer_end = 30
 forecast_probability = 70
 rp = 2
@@ -75,7 +75,7 @@ for code, station, in stations_adm2.items():
     ncols = round(nevents/nrows)
     igroup = 1 
     
-    fig, axs = plt.subplots(nrows, ncols, figsize=(10,10), squeeze=False, sharey=True)
+    fig, axs = plt.subplots(nrows, ncols, figsize=(20,20), squeeze=False, sharey=True)
     
     for i in range(nrows):
 
@@ -94,7 +94,7 @@ for code, station, in stations_adm2.items():
                 for ix, leadtime in enumerate(LEADTIMES):
                     glofas_reforecast_sel = (
                         ds_glofas_reforecast_summary[code]
-                        .shift(time=-leadtime) # TODO: I'm actually a bit unsure of whether this shift needs to happen or not...
+                        #.shift(time=-leadtime) # TODO: I'm actually a bit unsure of whether this shift needs to happen or not...
                         .sel(leadtime=leadtime)
                         .sel(percentile=forecast_probability)
                         .sel(time=slice(glofas_start_date, glofas_end_date))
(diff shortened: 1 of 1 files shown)


analyses/malawi/flooding_trigger/08_mwi_event_visualization.md
```diff
@@ -0,0 +1,130 @@
...
+
+                flood_onset_date = pd.to_datetime(events[station].loc[igroup-1, 'start_date'])
+                glofas_start_date = flood_onset_date - np.timedelta64(buffer_start, 'D')
+                glofas_end_date = flood_onset_date + np.timedelta64(buffer_end, 'D')
+                rp_val = df_return_period.loc[rp, code]
+
+                for ix, leadtime in enumerate(LEADTIMES):
+                    glofas_reforecast_sel = (
+                        ds_glofas_reforecast_summary[code]
+                        .shift(time=-leadtime) # TODO: I'm actually a bit unsure of whether this shift needs to happen or not...
```
turnerm, line 97: This can be removed! `utils.get_glofas_reforecast` shifts the dates by default so each date corresponds to whta is being forecasted for that date. (There is also the option to turn it off if needed)

analyses/malawi/flooding_trigger/08_mwi_event_visualization.md
```diff
@@ -0,0 +1,130 @@
...
+```python
+events = {}
+for station in stations_adm2.values():
+    events[station] = pd.read_csv(EXPLORE_DIR / f'all_rco_event_summary.csv')
+```
+
+Create a plot for each event. For the dates around each flood onset, we want to see what both the GloFAS reanalysis (historical) and reforecast data look like. We'll pick a specific forecast probability and include forecasts from various leadtimes. The y-axis is scaled to the percentage of a given return period threshold. Lines on the graph have a stronger weight when the values exceed the threshold.
+
+```python
+buffer_start = 5
```
turnerm, line 65: It might be good to increase this, since it's helpful to know if the forecast predicts something early, as that could be considered a true positive

turnerm (approved):
Really great visualization Hannah! I have a couple of ideas but they will definitely clutter up the nice look and are totally optional so feel free to ignore:
- Add a vertical line for the event date, especially if not using a symmetric before / after range
- Instead of making the forecasts all one colour, use a sequential colormap to denote leadtime

analyses/malawi/flooding_trigger/08_mwi_event_visualization.md
hannahker: Capture

So that's what I thought but was a bit uncertain with many of the figures looking like this without the shift in the forecast data... does this also look odd to you? (and like there's still a shift that needs to happen)

turnerm: You're right that this looks very suspicious (actually most of your examples do), I became a bit worried and did a few checks but now I'm sure that the code is working as intended, which is:
- `shift_dates = False` means that the date corresponds to the day the forecast was made
- `shift_dates = True` means that this level of discharge is forecasted for that date

The checks I did included doing some test downloads from the CDS interface and looking at the variable descriptions in the raw data. But I think one clear indicator that we have it correct is the following: the reforecast is run twice per week, and when you get the raw data, it's only populated for two days per week (rather than all dates), which indicates that the date corresponds to the one where the forecast was run. I hope that makes sense.

Another check that I found useful was to use the following code snippet in your notebook:
```python
# Slice time and get mean of ensemble members for simple plotting
start = '2001-01-15'
end = '2001-5-31'
station = 'G1724'

rf_list_slice = ds_glofas_reforecast.sel(time=slice(start,end))[station]
ra_slice = ds_glofas_reanalysis.sel(time=slice(start, end))[station]

rf_list_slice.median(axis=1).plot.line( x='time', add_legend=True, lw=1)
ra_slice.plot.line(label='Historical', c='k', lw=1)
plt.show()
```
If you try with `shift_dates=True` and `False` respectively, you get the following:
image
image

So while not shifting the peaks does give a better peak alignment right around 02-01, the rest looks a bit off, particularly once the discharge is decreasing, where I think the forecast should be performing better.

hannahker: Thanks for checking up on that! I'll remove the shift from this notebook. So do you think that the suspicious-looking lags in these plots from ```shift=true``` is just an artefact of the forecast?

turnerm: Yes, I was thinking it could be due to the sub sampling and subsequent interpolation, it may cause the peak to appear as if it is shifting every day. It could be something to ask Ervin though.

PR #151, opened 2021-08-12
rename folders Somalia and Ethiopia

Tinkaa:
Renaming the folders and a few ipynb's. 
Optimally we would also switch from ipynb to md but couldn't get to completely run the ipynb's quickly. So decided that is not priority and should be fixed once we work on those countries again.


hannahker (approved):
Looks good to me and agree about re-prioritizing for another time

Issue #148, opened 2021-08-06
READMEs for each country


hannahker:
Each country within the ```analysis``` directory should have an updated README. [This template](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/develop/analyses/country_template/README.md) can be used to get started. 

We currently have READMEs for: 
- Bangladesh

We need READMEs for: 

- [x] Burkina Faso
- [x] Chad
- [x] Malawi (to be expanded)
- [x] Nepal
- [x] Somalia
- [x] South Sudan
- [x] Ethiopia
- [x] Niger

hannahker:
@Tinkaa and @turnerm I think this makes sense as a first step in improving overall documentation. What are current thoughts on the template that I linked above? I proposed this a while ago so it might be a bit out of date. Curious to hear if you think that any sections are overkill, or if there's anything significant missing. It can also be a super rough starting point with lots of room for flexibility with the different context for each country.

turnerm:
@hannahker thanks so much for organizing this, I think the template looks good, definitely a great starting point as you said. Maybe I will have some comments on it after using it but I think I'll give it a go with Nepal (and probably an update for Bangladesh).

Tinkaa:
Yes thank you Hannah! And sounds good. I added Ethiopia to the todo's as well cause just saw it is in a horrible state. 
I liked @joseepoirier's idea of adding a status tracking to the pilots (Kanban style). 
What do you think about that and @joseepoirier do you think the Readme's would be a good place to incorporate that?

joseepoirier:
I think the Kanban idea came up when Leonardo was suggesting we move away from Trello (per-project cards) to a more status-based way (to do, in progress, done) of tracking our work. I don't know if it's a meaningful distinction, but the Kanban would therefore reflect the work we have to do / are doing / have done, rather than having a Kanban tracking the status of each pilot individually.

If we go that route, we could absolutely add a dated "Current Status" line in the README that indicates whether the pilot's trigger is under development, being monitored, or has activated.

Tinkaa:
Yes I think Kanban was more general but for the readme's I like the idea of a "current status". maybe monitoring vs activated requires too much updating. But at least whether the trigger is currently under development or not

hannahker:
@Tinkaa I agree with you there. What about 'Current status' with options including: 
- under development (for when we're actively working on it)
- completed (once the pilot is finished)
- on hold (pilot is still under development but analytical work is on hold)

[...]

PR #153, opened 2021-08-17
Add status badges to readme template

hannahker:
As discussed in #148

### DIFF
--- a/analyses/country_template/README.md
+++ b/analyses/country_template/README.md
@@ -1,6 +1,16 @@
 # [COUNTRY NAME] Anticipatory Action
 
-##  Table of Contents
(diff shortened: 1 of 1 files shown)


Tinkaa:
Great, thanks for updating!

Tinkaa:
@hannahker Oops I didn't look carefully. Small detail, but I would propose to change the ON HOLD to red such that we have a different color for each state. What would you say? 
`https://img.shields.io/badge/STATUS-ON%20HOLD-%23007CE0`
https://img.shields.io/badge/STATUS-ON%20HOLD-%23007CE0

hannahker:
Yes, good point! I'll just update this directly in ```develop``` if that's alright.

PR #154, opened 2021-08-18
Feature/readme bfa

Tinkaa:
What should have been one readme file change, became a change of 202 files😅
But this is only restructuring of files. I split them in commits. For the reviewing only use the `create readme` commit. 
Apart from that I
- Moved docs files to a docs folder
- numbered the md's
- removed old ipynb/html files
- removed the jupyterbook files

Questions: 
- Please add additional information on the vulnerability analysis in the readme as you see fit.
- What do you think of this folder structure? 
- Can you check if I didn't break your code with restructuring? 
- Would it be an idea to load the `ocha_logo.png` and `centre_logo.png` from our GDrive instead of saving them here?


hannahker:
Looks great! I think this folder structure is a really nice way to organize things. And good point on using hosted versions of the logos, rather than keeping them in a folder here. I put the logos in an `Assets/` dir within our `General - All AA Projects` drive folder. There could be an argument for moving that to one of the top-level directories in this repo, especially if we have other standard assets that we want to use for our docs. Although this is something that we can come back to as our documentation evolves. 

I noticed a couple outdated folder references in the readme and just fixed them myself since you'll be on holiday now.

hannahker (approved):
Looks great!

PR #158, opened 2021-08-18
update readme mwi

Tinkaa:
Update mwi readme

do you think it is detailed enough? 

Also, I would like to update the folder name to `mwi` instead of `malawi`, but we have shared the links to the summary documents (gh pages) with external partners. Would you know a solution to this?

### DIFF
--- a/analyses/malawi/README.md
+++ b/analyses/malawi/README.md
@@ -6,6 +6,42 @@ Two types of shocks to anticipate to are explored:
 
 ## Dry spells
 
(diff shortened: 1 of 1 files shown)


hannahker:
Yeah, that's a good question... I've had the same problem with broken links shared externally when we switched the name of the bangladesh directory. I don't suppose we've kept track of who has the link and could send an email update with the new link? Otherwise, I'm not sure of a great solution and we should ideally update the name to `mwi` sooner vs later.

hannahker (approved):

PR #160, opened 2021-08-18
Update/iri

Tinkaa:
Refactor the code to retrieve IRI's seasonal forecast. 
It was working but in an ugly state and have learned much since on geospatial data processing. 
Since I might not be the only one working with it anymore in the future, I hope it is now much better useable by others. 
The main file is `src/indicators/drought/iri_rainfallforecast.py`


src/indicators/drought/iri_rainfallforecast.py
```diff
@@ -43,44 +41,79 @@ def download_iri(iri_auth, config, chunk_size=128):
...
+    # used for but couldn't discover useful information in it and will
+    # give an error if trying to read both (cause C is also a variable
+    # in prob) the date format is formatted as months since 1960. In
+    # principle xarray can convert this type of data to datetime, but
+    # due to a wrong naming of the calendar variable it cannot do this
+    # automatically Thus first load with decode_times=False and then
+    # change the calendar variable and decode the months later
     iri_ds = xr.open_dataset(
-        IRI_filepath, decode_times=False, drop_variables="C"
+        input_path, decode_times=False, drop_variables="C"
```
hannahker, line 72: Given the experience with GloFAS data not opening properly with `xr.open_dataset()` maybe worth switching this to the same method that we've applied there? 

```
with xr.open_mfdataset(
                filepath_list,
                engine="cfgrib",
                backend_kwargs={
                    "indexpath": "",
                    "filter_by_keys": {"dataType": data_type},
                },
            ) as ds:
```

src/indicators/drought/iri_rainfallforecast.py
```diff
@@ -43,44 +41,79 @@ def download_iri(iri_auth, config, chunk_size=128):
...
     response = requests.get(config.IRI_URL, cookies=cookies, verify=False)
-    with open(IRI_filepath, "wb") as fd:
+    with open(output_path, "wb") as fd:
         for chunk in response.iter_content(chunk_size=chunk_size):
             fd.write(chunk)
 
-    # TODO: explore if can also open with rioxarray instead of xarray.
-    # Getting an error with current settings
+
+def clean_iri_data(
```
hannahker, line 49: A general comment, but have you been in touch with IRI about these data problems? Maybe they want to take on some of this processing internally to improve their data quality?

src/indicators/drought/iri_rainfallforecast.py
```diff
@@ -89,31 +122,9 @@ def get_iri_data(config, download=False):
...
 
-    # TODO: understand rasterio warnings "CPLE_AppDefined in No UNIDATA
-    # NC_GLOBAL:Conventions attribute" and "CPLE_AppDefined in No 1D
-    # variable is indexed by dimension C" Conventions attribute is
-    # caused by the fact that rasterio is expecting a conventions
-    # instead of Conventions. Shouldn't impact performance no 1D
-    # variable is indexed by C I am not entirely sure what is meant
-    with rasterio.open(IRI_filepath) as src:
-        transform = src.transform
+    iri_ds = xr.open_dataset(iri_filepath_clean)
```
hannahker, line 128: Same as comment above with using `xr.open_dataset()`

src/malawi/mwi_seasonalrainfallforecasts.py
```diff
@@ -1,3 +1,5 @@
+# TODO: this scripts needs serious refactoring
```
hannahker, line 1: Can add these kinds of comments to an issue so that we can keep better track or refactoring priorities?

hannahker (approved):

hannahker (commented):
Sorry for taking so long to get to this! I've also just looked through the code in `src` as I see there are merge conflicts with many of the notebooks. 

The only thing to maybe address/discuss before merging would be our method of opening files with xarray, which I've commented on where I noticed. I was having a lot of weird errors when opening the Glofas data with this method, which are fixed when we switched to the one that I commented.

src/indicators/drought/iri_rainfallforecast.py
Tinkaa: Good point! Think in this case should be `xr.load_dataset()` (as it is `.nc` and not `.grib`) so changed it to that

Tinkaa: btw big disadvantage is of `load_dataset` that it can take a long time to load with larger datasets. With IRI it is still alright but e.g. with chirps it is not so nice. But dont know of a better solution :(

src/malawi/mwi_seasonalrainfallforecasts.py
Tinkaa: Yes good one. I will make an issue of converting large todo's to issues
And for this file specifically, it is actually never used so deteleted it (easy fix :p )

src/indicators/drought/iri_rainfallforecast.py
Tinkaa: Hmm no, nice idea, I will

PR #162, opened 2021-08-31
Update Bangladesh README

turnerm:
Since we're cleaning up the READMEs, updated the Bangladesh one to better reflect the present situation

### DIFF
--- a/analyses/bgd/README.md
+++ b/analyses/bgd/README.md
@@ -2,6 +2,9 @@
 
 [![Generic badge](https://img.shields.io/badge/STATUS-ENDORSED-%231EBFB3)](https://shields.io/)
 
+Activated in July 2020, monitored in 2021.
+Last updated: Sep 2021
+
 ## Background information
 
 Work on the Bangladesh pilot began in early 2020. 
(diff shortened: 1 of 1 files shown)


analyses/bgd/README.md
```diff
@@ -1,33 +1,81 @@
 # Bangladesh Anticipatory Action
 
-## Background information
+[![Generic badge](https://img.shields.io/badge/STATUS-ENDORSED-%231EBFB3)](https://shields.io/)
```
hannahker, line 3: Could you add a 'Last updated' line to go along with this?

analyses/bgd/README.md
```diff
@@ -40,33 +88,65 @@ The content within this repository is structured as follows:
 │   └── config.yml             <- Parameters to configure the analysis
 │
 ├── trigger_development       <- Scripts related to the trigger analysis
-│
+│   ├── utils             
+│   │   └── utils.py          <- Utiltiy methods for the trigger development scripts and notebooks 
+│   ├── 01_get_glofas_data.py  <- [defunct] Download GloFAS data
```
hannahker, line 93: I'd argue that this level of detail here probably isn't necessary and might be tough to maintain. Could probably be fine letting each of the notebook titles speak for themselves?

hannahker (commented):
Just a couple relatively minor comments but looks great otherwise. Thanks for updating!

analyses/bgd/README.md
turnerm: Oops yes, done! And I also mentioned the activation

analyses/bgd/README.md
turnerm: Yeah good point, I'll take these out!

PR #163, opened 2021-09-02
Switched to pytest

turnerm:
Quick switch to Pytest `setup_class` method instead of using the `unittest.testcase` class.

However, @mcarans I was wondering if there could be an even more pytest-y way to do this using fixtures. I use the class setup to collect a few variables that are used throughout several of the tests. I don't have much experience with using fixtures so I thought that you might have an idea for this specific case.


turnerm:
Superseded by #179

PR #164, opened 2021-09-06
add isort

Tinkaa:
I thought to be smart and add isort to pre-commit to prevent unsorted imports
However, it would also put our beloved sys.path.append after all other imports. Which is not what we want because then it cannot correctly import the modules from `src`. 
Do you have an idea of how we could do this or shall we wait with isort after we have restructured? 

Example:
```
from src.indicators.drought.iri_rainfallforecast import get_iri_data

path_mod = f"{Path(os.path.dirname(os.path.realpath(__file__))).parents[1]}/"
sys.path.append(path_mod)
```


turnerm:
Thanks so much for looking into this Tinka. I gave it a quick try and it looks like `isort` preserves the `path_mod` code, so we can start using it! 

Note that when you run pre-commit on all the files it will also make some black and flake8 related changes, some of those were my fault because I hadn't yet properly set up pre-commit, sorry about that! 

I would put `isort` before `flake8` as it modifies the files. Also, I've read that it's useful to also include [seed-isort-config](https://pypi.org/project/seed-isort-config/) as well for determining which dependencies are third party. You can do that as follows:
```yaml
-   repo: https://github.com/asottile/seed-isort-config
    rev: v2.2.0
    hooks:
    -   id: seed-isort-config
 ```
This should be placed before `isort`.

**Comment by Tinkaa** [comment]
Huhhh I don't know how I thought path_mod was being messed up then, but great!
Also had to add some extra settings to make isort compatible with black. 
But actually lets rerun this isort after all the folder renaming cause this is giving merge conflicts. 
Can you check if you agree now with the `.pre-commit-config.yaml` and `.isort.cfg`? then I will add it to `develop` when we finished folder renaming

File: .pre-commit-config.yaml [hunk 1]
```diff
@@ -1,11 +1,20 @@
 #largely copied from https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/ and https://mitelman.engineering/blog/python-best-practice/automating-python-best-practices-for-a-new-project/
 repos:
+    - repo: https://github.com/asottile/seed-isort-config
```
turnerm, line 3: Oops actually it looks like my advice was outdated, see [here](https://pycqa.github.io/isort/docs/configuration/pre-commit.html)

.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
```
turnerm, line 1: Would it be possible to move this to the `pyproject.toml` file since we have it already?

.isort.cfg
```diff
@@ -0,0 +1,2 @@
...
+known_third_party = cartopy,cdsapi,cftime,coloredlogs,ee,geemap,geopandas,google,googleapiclient,indicators,matplotlib,netCDF4,numpy,pandas,pygsheets,rasterio,rasterstats,requests,rioxarray,scipy,seaborn,shapely,tqdm,utils,utils_evaluation,xarray,xskillscore,yaml
```
turnerm, line 2: Is this really needed, or can isort figure this out on its own? 

If we do need it, it would be good to split it up into multiple lines if possible, and also add a blank line at the end.

.isort.cfg
turnerm: Also, I didn't see anything here for black compatibility -- did you mean to include:
```
profile = "black"
```

.pre-commit-config.yaml
Tinkaa: Aha I just blindly copied without googling, sorry

.isort.cfg
Tinkaa: this was only because of the isort-config, so removed now

.isort.cfg
Tinkaa: oops yes put it here cause that was created by isort-config
Regarding black compatibility, I had that in the `.pre-commit-config.yaml`. But I like the idea of having it in `pyproject.toml` instead, so did that now

Tinkaa: Regarding order: I think it shouldn't matter. With the profiling the two should be compatible. If you don't do the profiling and run isort first, it is not working as black and isort will then undo each others changes and precommit doesnt like that

turnerm: > Regarding black compatibility, I had that in the .pre-commit-config.yaml.

Ah oops sorry I didn't spot it there, my bad

> Regarding order: I think it shouldn't matter. With the profiling the two should be compatible. If you don't do the profiling and run isort first, it is not working as black and isort will then undo each others changes and precommit doesnt like that

Sounds good!

turnerm (approved):

Tinkaa:
@caldwellst since you said you wanted to learn more about pre-commit etc, I added you as a reviewer. 
We worked through most of it already with Monica before you joined. 
The idea is to add one hook, named `isort` and this hook makes sure the imports are sorted according to PEP8 guidelines. 

It seems a lot of file changes, but the only relevant files are `.pre-commit-config.yaml` and `pyproject.toml`, all the other changes is the effect of running `isort` on all files. 

Let me know if you want to discuss or don't feel comfortable being the reviewer on this yet.

caldwellst (approved):
@Tinkaa Thanks for this, very interesting. Have done a lot of reading this morning on`.pre-commit-config.yaml` and `pyproject.toml` files and Flake8, Black, and isort. Did some local testing to see how it worked and also just triple check it all works and it does, yay! As well, usage matches the latest documentation I can find.

PR #165, opened 2021-09-06
Feature/mwi arc plot

hannahker:
Adding a simple notebook to make plots to compare ARC2 resolution against different admin levels in Malawi. This is going in the `exploration` sub-directory since it's not really critical to any analysis and we don't need to maintain this code in any way. I just wanted to make some quick plots and this code might be useful to hold on to. 

Looking for a quick sanity check on this.

### DIFF
--- a/analyses/malawi/dryspells_trigger/exploration/plot_resolution_comparison.md
+++ b/analyses/malawi/dryspells_trigger/exploration/plot_resolution_comparison.md
@@ -1,0 +1,77 @@
+## Plot ARC2 and CHIRPS resolution
+
+Simple notebook to investigate the resolution of ARC2 and CHIRPS data when compared against various admin levels in Malawi.
+
+```python
+import rioxarray
+from shapely.geometry import mapping
+import geopandas as gpd
+import xarray as xr
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+mpl.rcParams['figure.dpi'] = 300
+```
+
+Set the config values and parameters.
+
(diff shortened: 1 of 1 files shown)


analyses/malawi/dryspells_trigger/exploration/plot_resolution_comparison.md
```diff
@@ -0,0 +1,74 @@
...
+ADM3_SHP = RAW_DIR / config.SHAPEFILE_DIR / parameters['path_admin3_shp']
+
+PLOT_DIR = DATA_DIR / 'public' / 'processed' / 'mwi' / 'plots' / 'dry_spells' / 'arc2'
+```
+
+Read in data.
+
+```python
+# Admin boundaries
+df_adm3 = gpd.read_file(ADM1_SHP)
```
caldwellst, line 46: All of the code and outputs look good, I think just the adm3 and adm1 files got switched around so adm1 is being mapped instead of adm3 down below.

caldwellst (changes requested):
Just need to swap the `df_adm3` and `df_adm1` variable assignments.

analyses/malawi/dryspells_trigger/exploration/plot_resolution_comparison.md
hannahker: Good catch!

hannahker:
@caldwellst fixed the variables so just merging this now 👍

PR #166, opened 2021-09-07
rename malawi folders and usages

Tinkaa:
- Changed the folders named `malawi` in `src` and `analyses` to be named `mwi`
- Changed the files that used `malawi` as a parameter to use `mwi` instead. The uses are to retrieve the config file, which is now in `src/mwi` instead of `src/malawi`. 

This PR will break GH pages links but as discussed this is the way it is and the sooner we do it the better.


hannahker (commented):
Changes look good to me and thanks for doing this! 

It might be good to also update the links to the docs (`malawi` to `mwi` in the URLs) in the various READMEs where we have referenced them.

Tinkaa:
@hannahker good one, changed those. Will also change the reference on HDX once we merge.

PR #167, opened 2021-09-07
rename chad folders

Tinkaa:
In an effort to change all country folders to be named as their iso3 instead of their full name. 
the config file lives now in `src/tcd` instead of `src/chad` so changed those references.


joseepoirier:
All looks good. Would make sense to rename these two files, too (CHD seems to refer to Chad):

analyses/tcd/results/food_insecurity/CHD_historical_IPC_FN_CS_perc.png
analyses/tcd/results/food_insecurity/CHD_historical_IPC_FN_CS_pop.png

Otherwise ready to merge.

PR #168, opened 2021-09-07
rename nepal folder and usages

Tinkaa:
In an effort to change all country folders to be named as their iso3 instead of their full name.
the config file lives now in `src/npl` instead of `src/nepal` so changed those references.


turnerm (approved):
Thanks for taking care of this Tinka!

PR #174, opened 2021-09-08
fix saving path

Tinkaa:
@caldwellst @hannahker super small change to update an old path for saving. So just for your awareness. @caldwellst feel free to merge directly.


caldwellst (changes requested):
Looks good! I just think the directory on line 151 also needs to be changed accordingly.

Tinkaa:
Good catch, thanks!

PR #175, opened 2021-09-08
Add ADM1 boundaries to plots

caldwellst:
Very simple, just wanted to add the ADM1 boundaries to the plots so I could better understand the regions (and in particular easily identify the southern region when viewing the plots).

Edit: also added ADM2 boundaries to refer to for my better understanding of current documentation.


hannahker (approved):
Looks good to me!

PR #177, opened 2021-09-08
added ner-readme 

joseepoirier:
issue #148


analyses/ner/config.yml
```diff
@@ -0,0 +1,25 @@
+#general country information
```
Tinkaa, line 1: We don't have configs in the `analyses` folder, only in the `src`. So this can be removed imo. 
Aah I now see that is in the template, sorry! That has to be updated, will do that tomorrow.

Tinkaa (changes requested):
Thanks Josée! I think the information is good. 
My comments on removing sections are only to make it consistent with the other readme's.
And I am sorry that I think you copied from the template folder, but I now see that that is quite outdated. I will update that. 
But I would suggest to also remove the folders within the `ner` folder you created, such that `analyses/ner` only has a `README.md`. The reasoning being that it is redundant.

analyses/ner/README.md
```diff
@@ -0,0 +1,45 @@
...
+1. [Background Information](#background-information)
+2. [Overview of analysis](#overview-of-analysis)
+3. [Data description](#data-description)
+4. [Directory structure](#directory-structure)
+5. [Reproducing this analysis](#reproducing-this-analysis)
+
+## Background information
+
+Anticipatory action in Niger aims to preemptively respond to drought. The trigger was defined to reflect a 1 in 3 year event. It is evaluated once monthly between January and June, inclusively. It is monitored by IRI/WFP (through their tool Maproom) and uses analysis developed specifically for this purpose.
+
```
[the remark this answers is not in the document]
joseepoirier: There are no publicly available documents on Niger. I didn't include the one to Maproom either as IRI is in the process of pushing a V2 and is changing the URL (also I'm not sure it is intended to be publicly shared.)

analyses/ner/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Overview of analysis
+
+CHD provided guidance but did not perform data analysis for this project.
```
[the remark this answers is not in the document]
joseepoirier: removed the `data description` section as well.

analyses/ner/config.yml
joseepoirier: @hannahker pointed me to the template folder. I take note the template is outdated (Hannah: FYI).

hannahker: Thanks for flagging this. Have just updated the template and making a PR now!

Tinkaa:
Thanks for the updates @joseepoirier !

Tinkaa (approved):

PR #178, opened 2021-09-08
Update country template

hannahker:
- Removed sections from the template README that we likely won't use
- Removed outdated default folder structure
- Added a template `.Rmd` file in the `docs` folder


analyses/country_template/docs/template.Rmd
```diff
@@ -0,0 +1,51 @@
+---
+title: '<img src="https://drive.google.com/uc?id=118y5T73-iSUZoAMtoJGddxq9QzD_GDKX" style="height:40px;float:left; margin:10px" /><img src="https://drive.google.com/uc?id=1fHQUzF3ZjaoHj9KQ33-94dK_X1hcmjzW" style="height:50px;float:left; margin:10px" />'
+pagetitle: 'Docs Template'
```
joseepoirier, line 3: Should we include comments to call out the lines in the header that need to be updated when the template is used?

analyses/country_template/docs/template.Rmd
```diff
@@ -0,0 +1,51 @@
...
+<br>
+
+```{r setup, include = FALSE, message = FALSE, warning = FALSE}
+# source("") # Link to external script
+library(kableExtra)
+library(knitr)
+knitr::opts_chunk$set(echo = FALSE) # do not print code by default
+knitr::opts_chunk$set(include = FALSE) # do not print output by default
+```
+
```
joseepoirier, line 33: Maybe we add the variable assignment for the path to the Data folder?

analyses/country_template/docs/template.Rmd
```diff
@@ -0,0 +1,51 @@
...
+<br> 
+<br>
+<br>
+
+# Documentation Template
+Analysis last updated `r Sys.Date()`
+
+<br>
+
```
joseepoirier, line 25: Do you think it'd be helpful to add a section on global options (remove scientific notations, do not print output of chunks, etc.)?

Also, do we want to follow the practice of loading an analysis script (.R) or reproducing parts of a full analysis in an Rmd? What about functions: should they be defined in a separate script?

analyses/country_template/docs/template.Rmd
```diff
@@ -0,0 +1,51 @@
...
+```{r setup, include = FALSE, message = FALSE, warning = FALSE}
+# source("") # Link to external script
+library(kableExtra)
+library(knitr)
+knitr::opts_chunk$set(echo = FALSE) # do not print code by default
```
joseepoirier, line 30: That answers my comment above :)

analyses/country_template/docs/template.Rmd
```diff
@@ -0,0 +1,51 @@
...
+knitr::opts_chunk$set(include = FALSE) # do not print output by default
+```
+
+
+Copy this basic template to get started with creating an R Markdown file to document an analysis.
+
```
joseepoirier, line 36: Do we want to provide a TOC/section template? Like Background, etc.? I don't see it as necessary but if we bother creating a template, would be worth providing some structure so similar info is documented between analyses.

analyses/country_template/docs/template.Rmd
```diff
@@ -0,0 +1,51 @@
...
+
+```{r maps, dpi=300, include=T, message= FALSE, warning=FALSE, fig.height=8, fig.align="center"}
+
+# Put a figure in here, for example 
+
+```
+
+
+## Heading 1
```
joseepoirier, line 45: I don't think Headers are necessary - a RMarkdown cheatsheet provides that info.

joseepoirier (commented):
Looks good to go as is. I wonder what additional info/structure may make this template more useful.

analyses/country_template/docs/template.Rmd
hannahker: Yes, good idea

analyses/country_template/docs/template.Rmd
hannahker: I think we could probably assume that this might be in a linked analysis script?

analyses/country_template/docs/template.Rmd
hannahker: Yeah, good call 👍

analyses/country_template/docs/template.Rmd
hannahker: I'd say that we probably don't yet have enough consistency in what goes into these docs to make that kind of template meaningful yet. Hopefully this can develop further as we use this kind of documentation more

analyses/country_template/docs/template.Rmd
joseepoirier: If we think Rmd's will only be used to follow up on / build on external scripts than I think it's okay.

PR #179, opened 2021-09-09
Test GloFAS processing step

turnerm:


mcarans:
Sorry that I have taken so long to look at this PR. I need access to the Google Drive folder (and have requested it) in order to download the data needed to run the test (AA_DATA_DIR).

On your questions:
pytest does not require the use of classes. You can just use functions. However, I'd suggest that if there are multiple functions, that they be grouped in a class. 

Something that may be useful is that pytest allows for a file called conftest.py that can contain fixtures which are shared between different sets of tests. It can be placed at the root of the tests folder and fixtures in there are accessible to all tests.

I have utilities to deal with temporary folders. See: https://github.com/OCHA-DAP/hdx-python-utilities#path-utilities

Specifically I'd suggest something like the codebelow. The temporary folder AATest is created inside the tmp folder (or the folder defined in the environment variable TEMP_DIR if you have set it). If it already exists, it is recreated. Since the scope of the fixture is class, it is called once on class setup (so the same temp_dir can be used in all functions in the class). If the tests exit successfully AATest is deleted. If not, it remains (which can be helpful for debugging test failures).

```
@pytest.fixture(scope="class", autouse=True)
def temp_dir(self):
    tempdir = get_temp_dir("AATest", delete_if_exists=True)
    yield tempdir
    rmtree(tempdir)
```

mcarans:
I've made the tests use pytest fixtures and pytest-mock and cleaned things up a bit so there is no dependency on the existence of the environment variable AA_DATA_DIR or on .cdsapirc in the home folder to run the tests.

mcarans (approved):
I've made changes to address the points you made so it's back to you to review my changes :-)

Issue #173, opened 2021-09-07
Propose geographic threshold for trigger


hannahker:
We need to consider the number of admin regions that are required to be experiencing a dry spell for the trigger to be met. This parameter should be based on a historical analysis of the frequency of occurrence of various options so that we can propose a trigger for an event that is sufficiently extreme (for example, the even should be more extreme than having a likelihood of occurring every other year).

hannahker:
@joseepoirier in thinking through this issue I've realized that I could use clarification on a couple points relating to how the trigger would be acted on from an operational perspective, given that we are effectively going to be monitoring a number of different locations: 

- Would an action plan for the trigger be developed at the same scale that we're monitoring (eg. admin 2 in this case), or would actions be taken throughout the entire Southern region?
- If actions are specific to an admin 2, would it theoretically be possible to trigger multiple times within the same monitoring period, but for different admin 2 regions?

joseepoirier:
@hannahker A trigger is always used to decide WHEN to act, not WHERE actions are implemented. Using better spatial resolution improves our ability to detect more localised shocks and provides information to partners as they plan their activities.
For your second point: let's talk this through.

caldwellst:
Interesting question/response, look forward to discussing! A few other questions from my side as well:

### Dry spell severity

When we discuss severity of a dry spell, is it more severity based on wider geographic spread or length of the dry spell? It seems there were just a few instances of intensely prolonged dry spells in the analysis document (up to around 20+ days or so). Are these "more severe", or is the 14 day period when most damage is already done and there's diminishing marginal impact for additional days afterward? 

### Admin2 vs. raster?

Looking at the graphs produced by Hannah of ARC2 resolution at the admin 2 level, I was thinking about why not to just directly use the rasters themselves for measuring geographical coverage of a dry spell. Given the lower resolution of ARC2, this might be more accurate. For instance, Zomba City looks to comprise only a single centroid and thus has 3 recorded dry spells whereas Zomba itself has none.

<img width="651" alt="image" src="https://user-images.githubusercontent.com/8086547/132521025-0896c67f-4c49-410a-93fe-3601d4573f09.png">

As well, just anecdotally from the data presented above, you could see a case where there might be a large swath of area in dry spell from northern Chikwawa through Thyolo to Mulanje, however it looks unlikely that any of those specifically would register as a dry spell at the ADM2 level.

[...]

PR #181, opened 2021-09-13
Feature/obs trigger

hannahker:
Calculating basic statistics to address #173. I've looked at the frequency of triggering over the last 20 years across various thresholds for the minimum number of admin 2s needed to be experiencing a dry spell. I've calculated statistics for two potential monitoring situations: 

1. We're requiring dry spells across admin 2s to be identified *on the same day* for them to be counted towards the threshold, or
2. Dry spells across admin 2s can be counted cumulatively *within a given buffer of time* (here I set to 1 week)

I'm also aware that the underlying ground truth dataset of ARC2 dry spells may need to be updated as some of the admin 2s don't have any data when aggregating using the pixel centroid approach. 

As the main figure below shows, the results so far seem to be suggesting that we might expect to trigger once every 5 years if we required >= 3 admin 2s to experience a dry spell, each within 1 week of each other. @joseepoirier would this be an appropriate frequency?

Capture


caldwellst:
This looks great @hannahker! I was wondering what this would look like if we just had the trigger set on # of dry spell events triggered throughout that observational period at all?

joseepoirier:
@hannahker Thanks for putting this together! To confirm I understood correctly: is `single` referring to having X ADM2's experiencing on >= 1 day in common vs `cumulative` referring to X adm2's experiencing dry spells with at the same time or within 7 days of each other? Did you require a minimum overlapping duration?

I think a 1 in 5 year event would be an appropriate frequency.

hannahker:
@joseepoirier I've looked at the dry spell confirmation date since this is I think most relevant to our monitoring scenario. So: 

- `single` = we're looking for when X ADM2's have a dry spell confirmed on the same date
- `cumulative` = we're looking for when X ADM2's have dry spells confirmed within Y days of each other.

caldwellst:
@hannahker The newly updated dry spell detection files are available at `public/exploration/mwi/arc2/`. They are:

- `mwi_arc2_centroid_dry_spells.csv`
- `mwi_arc2_touching_dry_spells.csv`

However, turns out that taking the rolling sum of the rasters and then downsampling is methodologically identical to downsampling and then taking the rolling sum! Fun stuff.

hannahker:
@caldwellst to your question above, here is the plot with results when I adjust the 'buffer' around confirmations to extend to the whole monitoring period

Capture

I think the `Cumulative` line here (where the changes are reflected) is a little bit misleading though at the lower thresholds because it's not reflecting the fact we're seeing dry spells throughout a season across a number of admin regions, and this would involve triggering to respond to a dry spell in an admin region when there are still other dry spells potentially occurring days or weeks later in other admin regions, which this trigger would miss...

caldwellst:
Hi @hannahker, was thinking about this and playing around with the data trying to visualise it. However, I think simplest is to just look at the table of total dry spells confirmed that you produced.

```
date.          ADM2     Year 
2005-02-11	1	2005
2005-02-12	4	2005
2005-02-13	1	2005
2008-02-22	2	2008
2008-02-23	6	2008
2010-01-13	11	2010
2011-02-23	1	2011
2011-02-24	3	2011
2016-01-09	1	2016
2016-02-14	1	2016
2018-01-09	1	2018
2018-01-10	1	2018
2018-01-31	1	2018
```

I think the graph above is misleading as it shows all possible triggers within a year, and for single event trigger that counts multiple but for the cumulative does not. However, whatever the method, the trigger is only going to be activated once, so I think it makes more sense to analyse based on number of seasons we would meet the trigger.

For `n=1`, all trigger methods are actually going to have identical trigger date, the first time that any ADM2 area has a dry spell. There is no more triggering after that instance for any method so all of them would similarly miss later events I think.

If we set the limits to `n=2`, again, the triggers would be on the same day in all years except 2016, where the only method that would have the activation trigger be met would be cumulative. If `n=3`, again, all methods trigger on the same date except the only method to trigger in 2018 would be the cumulative method.

Based on how the trigger would work in practice, I think the risk of missing dry spells occurring later and not triggering occurs if we don't use the cumulative trigger. And looking at frequency, it's particularly interesting that if we set `n=3` and didn't use the cumulative trigger, there would've been no activation since 2011.

From another lens, it's the case that our choice of trigger (unless we went `n > 4`) is just deciding whether or not to activate in 2016 and 2018. Clearly there is no easy *scientific* decision to be made on such sparse data and a lack of impact data to validate against, but my intuition would be that having a trigger that misses 2018 (and thus any year since 2011) is a high risk that meets overall frequency criteria but might miss a shift in frequency in the past 10 years (total conjecture of course). Interested to get your thoughts!

hannahker:
@caldwellst these are really good points and I agree about the graph not being the most informative or accurate. Thankfully (?) the list of activations is small enough that the best way to present this is probably through the table, as you say. I've updated our slides with some of the trigger options that we've discussed and shown where activations would have happened or not. 

@joseepoirier can weigh in here on the extent of our decision-making responsibility, but the best that we can do might be to present a couple of multiple options and ask the field team to pick what they think is most appropriate.

hannahker:
@caldwellst let me know if you're happy for this to be merged!

caldwellst:
@hannahker @joseepoirier 

Okay, I put in a PR with some quite messy code I did to quickly explore all possible dry spells detected until March 13, and eventually compare that with the onset of the dry season. You can see more of the mess in the push I made (apologies for dirtying up that file haha!), but I've just put the final table and my word vomit below for your quick review!

pcode | dry_spell_confirmation | dry_spell_first_date | dry_spell_last_date | dry_spell_duration | overlap_dry_season | dry_season_first_date | no_year
-- | -- | -- | -- | -- | -- | -- | --
MW310 | 2002-03-04 | 2002-02-19 | 2002-03-04 | 14 | False | 2002-03-15 | 03-04
MW315 | 2005-03-13 | 2005-02-28 | 2005-04-17 | 49 | True | 2005-03-15 | 03-13
MW313 | 2005-03-13 | 2005-02-28 | 2005-04-17 | 49 | True | 2005-03-15 | 03-13
MW304 | 2005-03-13 | 2005-02-28 | 2005-04-17 | 49 | True | 2005-03-15 | 03-13
MW307 | 2005-03-13 | 2005-02-28 | 2005-03-18 | 19 | True | 2005-03-15 | 03-13
MW306 | 2005-03-13 | 2005-02-28 | 2005-04-17 | 49 | True | 2005-03-15 | 03-13
MW312 | 2005-03-14 | 2005-03-01 | 2005-04-17 | 48 | True | 2005-03-15 | 03-14
MW305 | 2005-03-14 | 2005-03-01 | 2005-04-17 | 48 | True | 2005-03-15 | 03-14
MW309 | 2011-03-07 | 2011-02-22 | 2011-03-07 | 14 | False | 2011-03-25 | 03-07
MW306 | 2020-03-04 | 2020-02-20 | 2020-03-11 | 21 | False | 2020-03-15 | 03-04
MW313 | 2020-03-04 | 2020-02-20 | 2020-03-23 | 33 | True | 2020-03-15 | 03-04
MW315 | 2020-03-09 | 2020-02-25 | 2020-03-11 | 16 | False | 2020-03-15 | 03-09
MW310 | 2020-03-09 | 2020-02-25 | 2020-10-10 | 229 | True | 2020-03-15 | 03-09
MW311 | 2020-03-09 | 2020-02-25 | 2020-04-14 | 50 | True | 2020-03-15 | 03-09
MW308 | 2020-03-10 | 2020-02-26 | 2020-03-10 | 14 | False | 2020-03-15 | 03-10
MW305 | 2020-03-10 | 2020-02-26 | 2020-03-23 | 27 | True | 2020-03-15 | 03-10
MW304 | 2020-03-10 | 2020-02-26 | 2020-03-11 | 15 | False | 2020-03-15 | 03-10
MW307 | 2020-03-10 | 2020-02-26 | 2020-03-16 | 20 | True | 2020-03-15 | 03-10

### Seth word vomit

From above, we can see that many of these dry spell activations into March overlap with the beginning of the dry season, as it would be defined using ARC2. Often, this occurs when the confirmation occurs past the 7th of March, so that more than half of the dry spell is taking place within March. This is the case in 2005, where all dry spells overlap with the dry season onset and in 2020, where 5 of 9 detected dry spells overlap with the onset of the dry season. The other 4 did not overlap with the dry season, although there was only a few day gap between the cessation of the dry spell and dry season onset.

The decision here really comes down to how we view these dry spells activated in 2020. Even ignoring dry season overlap, for all other years, there would be no change in activation based on these additional dry spells (there would be only 1 detected in 2002 and the trigger would've already been hit in 2005 and 2011).

So, how do we view the 2020 dry spells? Some general thoughts are:

- If we set a limit that at least half of the dry spell must occur in February (detect on or before March 7th), based on not wanting to overlap with the dry season onset, we would not have any additional activations based on our `>=3` trigger. We would still capture what appear to be legitimate dry spells in the rainy season in 2002 and 2011, while only capturing 2 dry spells in 2000. So it comes down to if we would have been okay missing the 2020 dry spells.

- For 2020, could we be conflating an early onset dry season with dry spells during the rainy season? I have seen this referenced in the literature as an issue (shortened rainy seasons), but is this something we want to capture with this trigger? The March 15th reference for rainy season cessation comes for a paper nearly 20 years old now, so it might be out of date regarding the seasonality in Malawi? It looks fairly clear that in 2005, for instance, the dry spells detected at the tail end of the period (only 1 day following in February) are simply capturing the dry season onset.

- If this is simply capturing a shortened rainy season, we could consider for V2 looking into capturing when the rainy season begins and the possibility of the cessation of the rainy season early. For instance, continuing to monitor for dry spells following March 7th and if we detect a situation like in 2020, triggering something based on the detection of a shortened rainy season (depending on its onset that year). For this round, we just want to be comfortable NOT activating in 2020 even though there were 9 separate ADM2 areas with a dry spell detected.

My intuition is that we are okay not activating the trigger for 2020. Cursory searches for documents/news about Malawi dry spells in 2020 have produced no collaborating documentation on high impact dry spells in that year. While this is obviously quite ancedotal, it's quite easy to found references to dry spells and their potential impacts even in the year they are observed, such as 2005 or 2010. In fact, documentation from the time, such as this [April 2020 FEWS-NET report](https://reliefweb.int/sites/reliefweb.int/files/resources/MW_FSOU_April_2020_Final.pdf) references above average seasonal rainfall and subsequent crop production.

### Recommendation

I therefore would recommend that we DO include dry spells confirmed in March, but only monitoring until March 7th. Maintaining the `>=3` ADM2 trigger would keep the same return period and observed years where we would meet the action trigger, while also capturing some late onset dry spells without significant overlap into the dry season (only one instance in our observed data from 2020).

PR #182, opened 2021-09-13
Feature/mwi arc2 chirps compare

caldwellst:
Hey @hannahker, here's some of the initial work on comparing ARC2 and CHIRPS data. It primarily is composed of:

- `exploration/compare_arc_chirps.md` which I should rename, as was initially going to be comparison, but decided to use `.Rmd` for comparison and just left this for data processing. Will be expanded once we have new ARC2 raw data.
- `exploration/compare_arc_chirps.Rmd` where I'm going to put all the comparisons and its knitted `.html` file.
- `12_arc2_identify_dry_spells.md` simple change to downsample raw raster data rather than rolling sum.

I will be updating these once new data is available and can calculate based on the method used for CHIRPS, but you can find the two different ARC2 dry spell data (you're already using the centroid one):

- `public/exploration/mwi/arc2/mwi_arc2_dry_spells.csv`
- `public/exploration/mwi/arc2/mwi_arc2_touching_dry_spells.csv`

You can also find those same files filtered to CHIRPS dry spells, but not necessary for trigger comparison.


hannahker:
A quick note that you might want to merge in the main `develop` branch as we've recently refactored the `malawi` directory to `mwi`.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+  group_by_all() %>%
+  summarize(Occurrences = n(),
+            .groups = "drop") %>%
+  mutate(across(-Occurrences, ~ ifelse(.x, "Detected", "Not detected"))) %>%
+  arrange(desc(Occurrences)) %>%
+  knitr::kable()
+```
+
+We can see above a breakdown of dry spell detection by method. Interestingly, where
+CHIRPS detects a dry spells, both ARC2 methods either also detect one or don't.
```
hannahker, line 238: So this would be confirmation that both ARC2 methods are consistent with each other?

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+         touch_chirps = chirps_start - arc2_touch_start) %>%
+  arrange(touch_chirps) %>%
+  mutate(row_id = row_number()) %>%
+  pivot_longer(c(center_chirps, touch_chirps),
+               names_to = "arc2_method",
+               values_to = "day_diff") %>%
+  mutate(arc2_method = ifelse(arc2_method == "touch_chirps",
+                              "Centroid / Touching",
+                              "Centroid only")) %>%
+  ggplot(aes(x = row_id, y = day_diff, color = arc2_method)) +
```
hannahker, line 253: Each dot here corresponds to a dry spell, correct? And so am I understanding this correctly as cases where the dry spell detections don't exactly match between the two (so all of the 27 dry spells in the first row of the table above), but we see that rather than missing the dry spell, CHIRPS is just delayed in detecting it?

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+
+### Trigger activation
+
+Let's consider if we're using these datasets for trigger activation and how they would
+compare. We will only consider dry spells that are confirmed at some point in January
+and February. Let's just quickly look at the number of dry spells detected this time period
+for each method to start.
+
+```{r trigger-chirps}
+chirps_df %>%
```
hannahker, line 353: These are still throughout all of MWI, correct? ie. not filtered to just the southern region

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+  theme_bw() +
+  geom_abline(intercept = 0)
+```
+
+While it looks like the touching method might skew higher (if just slightly) at
+higher levels of rainfall, we are more concerned at their behavior at lower levels.
+
+```{r 2d-histo-arc-1}
+daily_df %>%
+  filter(arc2_centroid <= 1 | arc2_touch <= 1) %>%
```
hannahker, line 421: Sorry maybe I'm not totally following here, but how is it that the axes on the figure below extend past 1 if you're filtering for precipitation values <= 1?

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+
+And above we can see that the centroid method in general trends much higher at these levels.
+This intuitively makes sense as the touching method takes the average of more raster
+cells than the centroid method, so will tend to regress toward the mean at the extremes.
+Anyway, given their similar results in identifying dry spells and the results above,
+should be fine to simply compare the centroid method against CHIRPS and not the
+touching method.
+
+```{r 2d-histo-chirps}
+daily_df %>%
```
hannahker, line 437: Not critical, but what does it look like if you colour the pixels according to a log scale? Just a bit difficult to see differences in the z axis since it all looks the same colour.

hannahker (commented):
Have only reviewed the `.Rmd` but all there looks really great! Also I've learned some new ways to wrangle data in R 😄 

Would you agree with these main takeaways from your analysis? 

- Daily precipitation estimates from CHIRPS and ARC2 are generally in agreement in Malawi. However, both across all values and at low precipitation levels, CHIRPS has the tendency to estimate precipitation to be slightly higher. 
- The difference highlighted above means that ARC2 is actually more sensitive to dry spells. When comparing lists of historical dry spells from both sources, CHIRPS identifies 25% fewer dry spells than ARC2. While we don't have a 'true' list of historical dry spells to compare against, this means that ARC2 may actually be *better* for our dry spell monitoring than CHIRPS. 
- Of the two methods explored for aggregating ARC2 data (touching cells vs cell centroid), the centroid-based approach is probably better for our purposes. However, both methods are well-aligned in their dry spell detections. The key thing to be aware of with the centroid-based method is that Zomba City doesn't contain a cell centroid and so doesn't have any data.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+  ggplot(aes(x = arc2_centroid,
+             y = chirps)) +
+  geom_bin2d(bins = 200) +
+  theme_bw() +
+  geom_abline(intercept = 0)
+```
+
+```{r 2d-histo-chirps-2}
+daily_df %>%
```
hannahker, line 446: It might also be interesting to disaggregate the ARC2 vs CHIRPS comparison by time of year. So in addition to looking at low precipitation values, how well do ARC2 and CHIRPS agree in their estimates during our monitoring period (ie. Jan/Feb)?

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+
+With the 3 separate datasets loaded, we can use the `fuzzyjoin` package to join the
+datasets together (with some repititive filtering after) to capture all overlapping
+dry spells to examine in detail how the 3 separate methods for dry spell identification
+compare for each incident in particular.
+
+```{r wrangling}
+# joining ARC2 touch and center dry spells to get all overlaps
+
+arc2_center_lj1 <- fuzzy_full_join(
```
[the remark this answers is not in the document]
caldwellst: Essentially! Gets messy because joining 3 data frames together so might be ways to clean it up but couldn't figure out a cleaner way in such a short time frame.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
caldwellst: Yeah, at least in their difference with CHIRPS. Basically used that to say I was comfortable just comparing ARC2 centroid to CHIRPS following from there.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
caldwellst: Yeah exactly! I thought was important in case ARC2 for instance was delayed in detecting, if we might need to consider using CHIRPS data at least around the time of its release or something. But seems in all cases of overlap, CHIRPS detected dry spells either on time or with some delay compared to ARC2.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
caldwellst: Good point. Might compare just across the southern region further on.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
caldwellst: I'm filtering to instances where at least one of the measurements is `<= 1`, so those values are showing large deviations for a measurement compared to the other.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
caldwellst: Yeah good call.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
caldwellst: Great point, can look at that as well as the regional disaggregation to see how they compare.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
```diff
@@ -0,0 +1,483 @@
...
+to only occur during rainy seasons identified using CHIRPS data.
+
+Since the rainy season definitions would slightly vary using ARC2 definitions,
+some of the dry spells recorded captured in those datasets may be instead the
+onset of the dry season and not captured if we defined the rainy season using ARC2
+data. After some exploration, I found that for ARC2 dry spells with a duration `> 30`,
+for both the centroids and touching method, only a single one of the
+20 and 18 dry spells respectively overlapped with a CHIRPS identified dry spell, and
+none had a dry spell confirmation in January or February. Thus, I explicitly filter
+these out prior to comparing the various dry spell identification methods.
```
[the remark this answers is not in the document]
caldwellst: Will make more explicit, it's the `20` and `18` reflected above.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
caldwellst: Added in update and even without explicitly filtering by region, the only confirmed triggers in those months were in the southern region.

analyses/malawi/dryspells_trigger/exploration/compare_arc2_chirps.Rmd
hannahker: Ah I see, so where ARC2 would be 0.2 but CHIRPS could be something like 100, for example?

caldwellst:
Okay, have gone through all the data and analysis to confirm that the joins between datasets is correct. Will put all relevant information into the Powerpoint.

hannahker (approved):
Looks good to merge to me 👍

PR #184, opened 2021-09-14
remove logos

Tinkaa:
Proposing to remove the logo files here as 
1) they are currently not being used anywhere
2) they are saved on our GDrive which imo is a better place for them. The logos saved there are e.g. used in `analyses/country_template/docs/template.Rmd`


joseepoirier:
Been wanting to do this, thanks for taking the lead! I thought some Rmd's include the logos (COVID reports + >= 1 risk analysis). It's an easy fix if they do and we ever need to re-run those scripts.

Tinkaa:
Perfect :) 
I checked if there were any references to them but couldn't find any. But yes we can replace easily and copy the GDrive link that is now used in some of the Rmd's

PR #185, opened 2021-09-14
Feature/mwi version

Tinkaa:
We changed the `dry_spells` data dir to have a `v1` and `v2` and within the `v1` an `archive`. 

This PR changes the dirs in the script to use the correct folders

Kind of boring PR :( I tested most scripts and in the worst case that something breaks, it is not that bad


caldwellst (approved):
All looks good!

PR #186, opened 2021-09-15
Remove dashboard from repo

caldwellst:
Removes the dashboard from the repository. I think we should consider completely removing the dashboard AND history as I've shifted the files with history to a [new repo](https://github.com/OCHA-DAP/pa-dashboard). However, wouldn't do that until we can shift the shinyapps.io deployment to the new repository.


caldwellst:
Okay, dashboard redeployed from new repo with no issues.

Tinkaa:
Awesome, looks good to me
Yes might be nice to do at some point, but there is no sensitive information so think it is not very important. 
Thanks for doing this!

Tinkaa (approved):

caldwellst:
Not hard, left like that because I thought it was used for various countries and might change, but will change to `pa-dashboard-eth` for now!

PR #188, opened 2021-09-17
Food insecurity: small refactor

Tinkaa:
Changes the usage of country name to iso3 as we have renamed all the folders to the iso3 codes. 
Also does some small other changes such as rounding numbers and adding the projection period. 

Note that the scripts are very messy and need a bigger round of refactoring still!

### DIFF
--- a/src/indicators/food_insecurity/ipc_definemetrics.py
+++ b/src/indicators/food_insecurity/ipc_definemetrics.py
@@ -1,4 +1,5 @@
 import logging
+
 import numpy as np
 
 logger = logging.getLogger(__name__)
--- a/src/som/som_foodinsec_trigger.py
+++ b/src/som/som_foodinsec_trigger.py
@@ -24,37 +24,42 @@ from src.indicators.food_insecurity.ipc_definemetrics import (
 )
 
 admin_level = 0
-country = "somalia"
+iso3 = "som"
 # suffix of filenames
-suffix = ""
+suffix_fn = ""  # "202106"
+suffix_ipcg = ""
+suffix_output = ""  # "202106"
 config = Config()
-parameters = config.parameters(country)
+parameters = config.parameters(iso3)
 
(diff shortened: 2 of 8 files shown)


src/som/som_foodinsec_trigger.py
```diff
@@ -57,14 +57,18 @@
     parameters[f"path_admin{admin_level}_shp"],
 )
 
+output_dir = os.path.join(country_data_processed_dir, "trigger_metrics")
+output_path = os.path.join(
+    output_dir, f"{iso3}_foodinsec_trigger{suffix_output}.csv"
+)
+
 
 df_fn = pd.read_csv(os.path.join(fewsnet_dir, fewsnet_filename))
```
caldwellst, line 66: Not directly part of your code refactoring above, but `fewsnet_filename` is no longer valid. They are named `somalia_...` in the file drive now. I didn't want to change in case used elsewhere, but flagging here.

src/indicators/food_insecurity/ipc_definemetrics.py
```diff
@@ -1,4 +1,5 @@
 import logging
+
```
caldwellst, line 2: Can discard extraneous space.

caldwellst (changes requested):
Everything looks good in the various files! I figured that the best thing to do was just test re-running the `src/eth/eth_foodinsec_trigger.py` and `src/som/som_foddinsec_trigger.py` files. Just a couple of changes there, 1 in code and 1 just in the `Data` folder on Google Drive that we may want to resolve now prior to merging.

src/som/som_foodinsec_trigger.py
Tinkaa: thanks for noticing, changed the names on the drive. They are referenced to in a few notebooks but the change should be either automatically reflected due to the use of the config file, or they were already out of date so won't change those now

PR #190, opened 2021-09-17
ecmwf download - change coordinates

Tinkaa:
This PR does two things:
- It generalizes the functions to extract an area that is given to the CDS API. This code is now found in `src/utils_general/area.py`
- For ecmwf's seasonal forecast the code is adjusted such that the coordinates can be rounded correctly. Before the raw coordinates were given, but to download the data correctly the outer bounds of the downloaded area should be integer coordinates. With float coordinates the downloaded data is shifted, which is incorrect.


caldwellst (commented):
@Tinkaa Everything looks good and is running on my machine! However, looks like you have merge conflicts, likely due to the new hooks, so would run `isort` locally and see if those resolve them.

Tinkaa:
@turnerm wanted to notify you about this PR as I changed the area code that is inputed to glofas CDS call slightly. Will merge now, but flag if you would do something different!

PR #191, opened 2021-09-20
change year and month input

Tinkaa:
Small PR to 
- Enable input of `year_min` and `year_max` to `download()`. Before the years could only be changed in the `ecmwf_seasonal.py` script, but now any year_min and year_max can be given when download is called
- Enable a custom list of months to be downloaded. Before the whole year had to be downloaded. 
- Remove the argument `split_by_month` as we haven't used it and could cause confusion when using the custom list of month

This PR merges into the branch of PR #190 so better to review and merge that branch first.


caldwellst (approved):
Looks good to me! Mainly been looking through how you're the various classes beyond the small changes you've made here. Good stuff.

PR #192, opened 2021-09-20
change ipynb to md

Tinkaa:
Changing the ipynb's to md format in `som` and `eth` folder. 

Some notebooks don't run fully anymore due to old dir structures. Will try to fix them soon (or add a disclaimer in the readme)


caldwellst (changes requested):
I think fine to merge now, but if `som` and `eth` are not going to be actively developed (so this change isn't urgent), maybe best to wait to merge once either the paths have been fixed or explicit notes added to those notebooks.

Tinkaa:
Yeah actually I agree with you. I am just gonna leave them as ipynbs now, it is not that bad.

PR #193, opened 2021-09-21
Feature/mwi pred trig

Tinkaa:
Script to compute the predictive trigger for Malawi. 
It is very basic and not automated, but should only be run twice. 
I am planning to make tests for it later, and also probably some plotting function(s)
 
- I am struggling a bit with how we can best define the variables that are now global variables (i.e. in capitals). Should they go to `main()`? Or is there a better way to do that? (@turnerm tagging you here as well since we have been talking about this)
- @caldwellst I made a `trigger_metrics` folder. Do you think this structure and filenaming is suitable, also looking at the observational trigger? 

Also there was some problem with using incorrect coordinates when downloading the data. I haven't decided which version we will use for the monitoring yet, but implemented it such that it can easily be switched.


turnerm:
> I am struggling a bit with how we can best define the variables that are now global variables (i.e. in capitals). Should they go to main()? Or is there a better way to do that? (@turnerm tagging you here as well since we have been talking about this)

I think they are find for now, ideally these will be refactored to config in the (hopefully very near) future so I wouldn't sweat it too much

caldwellst:
The `trigger_metrics` folder looks good to me. I think regarding the global variables and config files is fine, but really think that a single one linear of file paths called through `os.path.join()` calling the raw strings rather than from config might be cleaner.

Tinkaa:
Thanks! Lets discuss how we can best define paths in a call :) Will merge for now

PR #196, opened 2021-09-28
Classify MDG ADM4 areas as urban or rural

caldwellst:
Merge code to classify MDG ADM4 areas as urban/rural with the rest of your Madagascar work. Data is saved in `public/processed/mdg/urban_classification`.


src/utils_general/ghs.py
```diff
@@ -0,0 +1,120 @@
...
+        "/GHS_POP_E2015_GLOBE_R2019A_54009_1K/V1-0/tiles/"
+        "GHS_POP_E2015_GLOBE_R2019A_54009_1K_V1_0_{column}_{row}.zip",
+    }
+
+    # Set filepaths
+    download_dir = os.path.join(output_dir, "ghs", "zip")
+    output_filepath = os.path.join(
+        output_dir,
+        "ghs",
+        "{country_iso3}_{ghs_type}_2015_1km_mosaic.tif".format(
```
Tinkaa, line 51: Is it correct that this data doesn't get updated? If so, would it make sense to use an `use_cache` boolean. Such that it doesn't automatically redownload data when data already exists? See e.g. [here](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/a1fefb4d34124393c87d50b0214887ae4dc36903/src/indicators/flooding/glofas/glofas.py#L87)

analyses/mdg/urban_exploration.md
```diff
@@ -0,0 +1,59 @@
...
+       (22,12)]
+
+# Download data
+
+get_ghs_data("SMOD", box, iso3, raw_dir)
+
+# Load data
+
+adm4 = gpd.read_file(adm4_path, crs='4326').to_crs('ESRI:54009')
+with rasterio.open(os.path.join(raw_dir, 'ghs', 'mdg_SMOD_2015_1km_mosaic.tif')) as src:
```
Tinkaa, line 43: Is there a reason you are using rasterio instead of rioxarray (or xarray)? I have generally found that rioxarray is a bit more reliable with managing coordinates and CRS's correctly (`rioxarray.open_rasterio()` could be used in this case)

analyses/mdg/urban_exploration.md
```diff
@@ -0,0 +1,59 @@
...
+    trans = src.transform
+
+cls = classify_urban_areas(adm4, smod, trans)
+adm4['urban_area'] = [x['urban_area'] for x in cls]
+```
+
+Save out results.
+
+```python
+adm4.drop('geometry', axis=1).to_csv(os.path.join(processed_dir, 'urban_classification', 'mdg_adm4_urban_classification.csv'))
```
Tinkaa, line 54: Just details, but I would generally comment out lines that save files to prevent accidental overwriting

src/utils_general/ghs.py
```diff
@@ -0,0 +1,120 @@
...
+    :param urban_min_class: Minimum raster value to consider as urban area.
+        Defaults to 21, corresponding to suburban or peri-urban grid cells.
+    :param urban_percent: Percent of raster cells classified as urban to
+        consider polygon as urban.
+    """
+
+    def urban_area(x):
+        return np.ma.mean(x >= urban_min_class) >= urban_percent
+
+    return zonal_stats(
```
Tinkaa, line 115: If you use (rio)xarray, you could also use our [own zonal stats function](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/a1fefb4d34124393c87d50b0214887ae4dc36903/src/utils_general/raster_manipulation.py#L110). As we have sometimes seen that `rasterstats.zonal_stats` doesn't handle the coordinates fully correctly. 
However, I trust it handles it correctly in this case so fine with leaving it as it is!

src/utils_general/ghs.py
```diff
@@ -0,0 +1,120 @@
...
+        polygons,
+        ghs_smod,
+        add_stats={"urban_area": urban_area},
```
Tinkaa, line 118: This is cool, learned something new :)

analyses/mdg/urban_exploration.md
```diff
@@ -0,0 +1,59 @@
...
+from src.indicators.drought.config import Config
+from src.utils_general.ghs import get_ghs_data
+from src.utils_general.ghs import classify_urban_areas
+
+# Setup
+
+iso3 = "mdg"
+raw_dir = os.path.join(os.environ["AA_DATA_DIR"], 'public', 'raw', iso3)
+processed_dir = os.path.join(os.environ["AA_DATA_DIR"], 'public', 'processed', iso3)
+adm4_path = os.path.join(raw_dir, 'cod_ab', 'mdg_admbnda_adm4_BNGRC_OCHA_20181031.shp')
```
Tinkaa, line 25: would be nice to define the admin level as a variable instead

src/mdg/config.yml
```diff
@@ -0,0 +1,13 @@
+iso2_code: mg
```
Tinkaa, line 1: just make sure to merge this file with `feature/plague_explore`'s version to enable merging

Tinkaa (approved):
Great stuff! 
A few comments to make it even a bit neater, but none are critical

I ran the `.md` to compute stats for adm3 and that worked perfectly :)

analyses/mdg/urban_exploration.md
caldwellst: Monica mentioned it when we discussed this afternoon. The initial code I was using from the COVID work I was using was built on `rasterio`, so I based on that, but could be worth changing in the future to rioxarray!

src/utils_general/ghs.py
caldwellst: That is a very good idea!

src/utils_general/ghs.py
caldwellst: Will do it this way in the future! And can amend this if we come back to it.

analyses/mdg/urban_exploration.md
caldwellst: Yes, done!

src/mdg/config.yml
caldwellst: Decided not to use so will actually remove from this branch.

PR #198, opened 2021-09-30
Feature/mwi arc2 hdx

caldwellst:
Code used to produce HDX upload files for ARC2 data, similar to what is available for CHIRPS currently on HDX. I have cleaned up the files slightly so they differ from CHIRPS a bit, so we can discuss how to align the 2 if needed.

The files are available at:

`public/processed/mwi/arc2/mwi_arc2_dry_spells_2000_2020.csv`
`public/processed/mwi/arc2/mwi_arc2_rainy_seasons_2000_2020.csv`

### DIFF
--- a/analyses/mwi/dryspells_trigger/15_mwi_hdx_processing.md
+++ b/analyses/mwi/dryspells_trigger/15_mwi_hdx_processing.md
@@ -38,20 +38,24 @@ rainy_merged_df = pd.merge(rainy_df, adm_df[['ADM2_PCODE', 'ADM2_32EN', 'ADM1_EN
                      'season':'season_approx',
                      'ADM1_EN':'region'})
                      
-rainy_merged_df['onset_date'] = pd.to_datetime(rainy_merged_df.rainy_season_onset)    
-rainy_merged_df['cessation_date'] = pd.to_datetime(rainy_merged_df.dry_season_first_date) - pd.to_timedelta(1,unit='d')
-rainy_merged_df['onset_month'] = rainy_merged_df.onset_date.dt.month
-rainy_merged_df['cessation_month'] = rainy_merged_df.cessation_date.dt.month
-rainy_merged_df['rainy_season_duration'] = (rainy_merged_df.cessation_date - rainy_merged_df.onset_date).dt.days
-
 # filter rows for seasons fully calculated
 rainy_filtered_df = rainy_merged_df[~rainy_merged_df.season_approx.isin(['1999-2000', '2020-2021'])]
+    
+rainy_filtered_df['onset_date'] = pd.to_datetime(rainy_filtered_df.rainy_season_onset)    
+rainy_filtered_df['cessation_date'] = pd.to_datetime(rainy_filtered_df.dry_season_first_date) - pd.to_timedelta(1,unit='d')
+rainy_filtered_df['onset_month'] = rainy_filtered_df.onset_date.dt.month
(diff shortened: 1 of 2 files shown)


analyses/mwi/dryspells_trigger/15_mwi_hdx_processing.md
```diff
@@ -0,0 +1,138 @@
...
+rainy_merged_df['rainy_season_duration'] = (rainy_merged_df.cessation_date - rainy_merged_df.onset_date).dt.days
+
+# filter rows for seasons fully calculated
+rainy_filtered_df = rainy_merged_df[~rainy_merged_df.season_approx.isin(['1999-2000', '2020-2021'])]
+
+# get precipitation for rainy season
+
+ij_df = pd.merge(rainy_filtered_df, daily_df, how='inner', on='ADM2_PCODE')
+ij_df = ij_df[(ij_df.date >= ij_df.onset_date) & (ij_df.date <= ij_df.cessation_date)].reset_index()
+prec_df = ij_df.groupby(['ADM2_PCODE', 'season_approx'])['mean_cell'].sum().reset_index().rename(columns={"mean_cell":"rainy_season_rainfall"})
```
Tinkaa, line 55: Pretty detailed, but would round the `rainy_season_rainfall` column to a few decimals for readability

Tinkaa: Also the `cessation_month` and `rainy_season_duration` somehow end up being floats, would make them ints

Tinkaa (commented):
Looks good! 

I think it would be good to align the two datasets, and where there are differences in column namings etc. yours are definitely better. 
So the differences I see are: 
- the `season_approx`/`season_name` column in the dry spells and rainy seasons file. I like how you include both years. It makes it a bit harder to process, but that is fine. Shall we change the CHIRPS to the same format as you made for ARC2? And rename the column to `season` instead of `season_approx` for both sources? 
- Then the `during_rainy_season` column you use True vs False whereas CHIRPS uses [0,1,NaN]. Lets change CHIRPS to True vs False
- For the rainy seasons CHIRPS file, the index and the `season_approx=2021` rows should be removed

For ARC2 the data from 2020-2021 season is not included right? How hard would it be to add that? Also the 1999-2000 season is missing in the rainy season file, shall we add that?

analyses/mwi/dryspells_trigger/15_mwi_hdx_processing.md
```diff
@@ -0,0 +1,138 @@
+## MWI processing for HDX
+
+Code to process MWI ARC2 rainy season and dry spell data frames for upload to HDX.
+
+```python
+import pandas as pd
+import os
+import fuzzyjoin
```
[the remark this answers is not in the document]
caldwellst: Whoops yes need to remove, I had accidentally put in mixing up the R library I like to use with a non-existent python library.

analyses/mwi/dryspells_trigger/15_mwi_hdx_processing.md
caldwellst: Will round. The other two are parsed as floats because there are `NaN` values and `int` types are not able to contain `NaN` apparently. Have moved filtration to be before calculations to correct this, although the resulting CSV is the same.

Tinkaa: To me it seems the csv is fine now :)

PR #200, opened 2021-09-30
Feature/mwi explore multiple ds

caldwellst:
Very quick double checking on the instances of 2 dry spells in a single season!

PR #206, opened 2021-10-08
Test ARC2 infilling

caldwellst:
Very simple testing of ARC2 infilling for #200.


Tinkaa:
Nice methodology! Looks good to me and as you said in the issue, think it is such a rare occurrence that it makes sense to use the interpolation

PR #208, opened 2021-10-12
improve histogram

Tinkaa:
Small PR to add a histogram, and update the leadtime from 2&4 to 3 months as this is what is used for the trigger now.

### DIFF
--- a/analyses/mwi/dryspells_trigger/09_ecmwf_monthly_skill_dryspells.md
+++ b/analyses/mwi/dryspells_trigger/09_ecmwf_monthly_skill_dryspells.md
@@ -286,17 +286,18 @@ df_for["season_approx"]=np.where(df_for.date.dt.month>=start_rainy_seas,df_for.d
 ```
 
 ```python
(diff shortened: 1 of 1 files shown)


analyses/mwi/dryspells_trigger/09_ecmwf_monthly_skill_dryspells.md
```diff
@@ -360,6 +361,23 @@ df_ds_for["month_abbr"]=df_ds_for.month.apply(lambda x: calendar.month_abbr[x])
 df_ds_for_labels=df_ds_for.replace({"dry_spell":{0:"no",1:"yes"}}).sort_values("dry_spell",ascending=True)
 ```
 
+```python
```
caldwellst, line 364: Could you add just a bit of context to what's going on here?

caldwellst (commented):
So, obviously there's lots that was there prior to this commit, just going through it really struggling to follow what's happening where, when, and why. Some minor comments just on the pieces pushed here would be helpful to be clear on what's happening and why.

PR #210, opened 2021-10-13
Feature/mwi example map

caldwellst:
Okay, so, I wanted to put together this PR to show the potential draft maps and also discuss some issues I see with them. To do this analysis, I looked at 2018, where there 3 dry spells and we would've activated the observational trigger.

First, we have a simple map I put together with rainfall across the monitoring period, starting on the 24th of December until the 31st of January, when we would have triggered the 3rd dry spell.

I also highlighted the districts that were triggered in that year. However, while doing the work, I began to think about the logic of creating and using a map like this and I would recommend that we actually don't proceed with these. My thoughts are:

* We don't actually want agencies to do targeting based on total rainfall. There could be sufficient rainfall in some areas and worse rainfall in others, but that isn't actually necessarily reflective of where dry spells have occurred and where targeted assistance would be needed.
* Although we might rely on the administrative boundaries for the triggering, it isn't actually that reflect of where we should be targeting given that the boundaries are relatively arbitrary. So, even if we instead just displayed, for instance, districts that experienced a dry spell, I think it is not as instructive as we think and we risk focusing on smaller districts (with fewer raster cells) more likely to experience a dry spell rather than the wider areas that have experienced one.

I think we should entirely avoid presenting this kind of map because it's liable to be used incorrectly in responding to where dry spells are. A simple idea would be rather than displaying the precipitation map as above, we just overlay raster cells, where for each raster we capture if it's experienced a dry spell in that period.

While I need to check my analysis, I think the map works as an example to show how valuable this method can be for identifying areas most impacted by dry spells, ignoring arbitrary admin boundaries.

We can update the maps based on comments and fine tuning for presentation tomorrow, for instance zooming in, changing text, etc., but I wanted to get these two examples up and discuss what we actually want to show before proceeding.

### DIFF
--- a/analyses/mwi/dryspells_trigger/09_ecmwf_monthly_skill_dryspells.md
+++ b/analyses/mwi/dryspells_trigger/09_ecmwf_monthly_skill_dryspells.md
@@ -23,6 +23,7 @@ import matplotlib.pyplot as plt
 import matplotlib as mpl
 import pandas as pd
 import numpy as np
+import math
 
 import seaborn as sns
 from mlxtend.evaluate import confusion_matrix
@@ -248,7 +249,7 @@ start_year=2000
 end_year=2020
 #just locking the date to keep the analysis the same even though data is added
 #might wanna delete again later
-end_date="5-1-2021"
+end_date="2-1-2020"
 ```
 
 ```python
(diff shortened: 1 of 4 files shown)


caldwellst:
After discussion with Josée, we're going to keep this open for discussion in the future, but for now just going forward with the precipitation maps. Added discussion from above to the repo. Will re-open in the future.

Tinkaa:
Don't know what you discussed, but my intuitition would definitely be to keep it at the raster level. However, I would show the total rainfall per raster cell instead of the binary dry spell or not (maybe having distinct coloring for <=2mm and above but then in several bins). Something like [CHC's CHIRPS maps](https://chc.ucsb.edu/monitoring/early-estimates/info)

PR #229, opened 2021-10-14
GloFAS test process 2

turnerm:
@mcarans Thanks so much for all the changes you made, I understand now why you did it yourself because it would have been a lot to explain! 

I'm still wrapping my head around these fixtures a bit but I think the code looks really great. I guess I have one question: with the way it was written previously, I had to explicitly mock e.g. [open_mf_dataset as imported in the glofas class](https://github.com/OCHA-DAP/pa-anticipatory-action/compare/f92ea69...426ed52#diff-ba7aff226ad826c86f223866724e1621906df439c4dfe8987563562267ee12d4R197).  With fixtures, are you mocking `open_mf_dataset` anywhere it's used by `xarray`? ([here](https://github.com/OCHA-DAP/pa-anticipatory-action/pull/229/files#diff-ba7aff226ad826c86f223866724e1621906df439c4dfe8987563562267ee12d4R296-R299)). Is that desired, what if you have two modules that use it and you only want to mock it for one?


turnerm:
Fixed in AnticiPy

PR #231, opened 2021-10-25
Feature/mwi update

hannahker:
Slight adjustments to the flood trigger analysis code to make sure that everything runs with updates to `src`. The only notable change is that with the updated `utils.get_detection_stats()` approach for finding a 'match' between glofas and a historical event based only on the start date, the detection stats are notably worse than the figures cited in the [summary slides](https://docs.google.com/presentation/d/17GUPKKzf-iU1WNmjV0bBeY1OXSjDmuDT0_Q_Rf0zkKw/edit#slide=id.ge325e9c230_0_82). :(

### DIFF
--- a/analyses/mwi/flooding_trigger/02_mwi_floodscan_events.md
+++ b/analyses/mwi/flooding_trigger/02_mwi_floodscan_events.md
@@ -11,6 +11,7 @@ import geopandas as gpd
 import numpy as np
 import pandas as pd
 import xarray as xr
+import rioxarray as rio
 from scipy import stats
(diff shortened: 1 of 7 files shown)


turnerm (approved):
Thanks Hannah, this all looks very clear! One thing though: would it be possible to preserve the "old" method of calculating the detection stats (even in an archive or scratch directory), so that we can reproduce the plots if needed? (Were they shared outside of our team?)

hannahker:
@turnerm good point! Have just added that in with the old functions.

PR #233, opened 2021-11-04
Feature/ecmwf unrounded

Tinkaa:
Small PR where I changed
- the wording from `incorrect` to `unrounded` as it is actually not incorrect to use unrounded coordinates, but the data is in that case interpolated. 
- changed the labels of two figures to be used in the technical note. 

I know the code is not so well-documented. I didn't put a great effort in improving it currently as it will be restructured again for the refactor, but if there is something immediate please comment.


caldwellst (approved):
Looks good to me!

Issue #234, opened 2021-11-05
Malawi map output for predictive trigger


Tinkaa:
Hi @joseepoirier @caldwellst ,
Related to #232 , we also got the request to create a map for the predictive trigger. I have been playing a bit with it and came up with the idea shown below. This map shows the forecasted monthly precipitation with a diverging scheme where the color changes at the threshold (=210 mm). 

I chose to display an area around Malawi as well, as it gives a bit more context. However, we could also choose to only show a smaller area or even only show the cells that are included in the computation of the trigger (which is the second image). 

However, I am struggling a little bit with what the map conveys and what it will be used for. 
In this case it might seem from the map that the trigger should be met. However, the trigger wasn't met in this case (ofc I chose the trickiest example). The reason for that is that only 2 cells are included in the computation, the ones shown in the second image, and in this case the mean of those two cells is larger than 210 mm. 

@joseepoirier @caldwellst what are your opinions? Should we produce a map and if so, is this a good format or do you have other ideas? 

To build upon the voting, I would currently vote for 
- Map with whole area
- No map (if this is an option)
- Map with only selected cells

![afbeelding](https://user-images.githubusercontent.com/12556634/140479303-5286f844-90ce-4e86-b118-deb7db4e238d.png)
![afbeelding](https://user-images.githubusercontent.com/12556634/140481372-31eb9905-e2e4-4f24-ad21-5a8f88183fe9.png)

joseepoirier:
Thanks Tinka. I believe @caldwellst has been working on this as well, so let's make sure to coordinate the efforts. This map must be ready by next week (ie code PR-ed and merged by Friday 12 Nov) as we may need to produce it as early as Monday 15 Nov. Please do loop in @milanoleonardo if needed during my absence.

The predictive trigger is based on the 50%+ likelihood of <210mm of rain in the Southern region (adm1). The email announcing it was met is to include the forecast for reference and transparency. The top map will convey the main message and provide some context around it too, without raising too many questions about the methodology (which are welcome, but not in response to an activation announcing email.)

caldwellst:
Thanks @Tinkaa, agree here with your voting and the suggestion by @joseepoirier. Since my assumption is that we are only producing this map and sending the e-mail if we reach the trigger, then I don't think we need to worry about confusing visualisations you raise.

joseepoirier:
[...]

PR #235, opened 2021-11-05
Feature/mwi pred map

Tinkaa:
Create a map showing the raster forecast, overlayed with the admin boundaries. This will be shared with the team if the predictive trigger is met, as discussed in #234. 

Also a few small general processing changes to make the computation of the trigger easier. 

I added a few questions (indicated by `question:`) for the map that would be great if you could give your input on. That is the only reason I am marking this as a draft PR for now.

### DIFF
--- a/src/indicators/drought/ecmwf_seasonal/ecmwf_seasonal.py
+++ b/src/indicators/drought/ecmwf_seasonal/ecmwf_seasonal.py
@@ -82,7 +82,7 @@ class EcmwfSeasonal:
             logger.debug(
                 f"{filepath} already exists and cache is set to True, skipping"
             )
-            return filepath
+            return False
         Path(filepath.parent).mkdir(parents=True, exist_ok=True)
         logger.debug(f"Querying for {filepath}...")
         query_str = self._get_query(
@@ -107,7 +107,7 @@ class EcmwfSeasonal:
         logger.debug(f"...successfully downloaded {filepath}")
         # Wait 2 seconds between requests or else API hangs
         time.sleep(2)
-        return filepath
+        return True
 
     def _get_raw_filepath(
         self,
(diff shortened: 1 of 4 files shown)


caldwellst:
Hey Tinka, so, regarding the questions:

### Graph text

I think that you should include the `"with {prob}% probability"` in the graph. However, I see your point about size. Could you instead put the date of publication as a caption using `figtext` and slightly larger margins around the bottom axis? Then you could use the newline to split up the plot title as necessary.

### Global constants versus function arguments

I think it's fine to have global constants, I just am not sure about calling global variables from within the function scope, because makes it hard to see exactly what is being called. There is probably a space and time for that, and good reason to use globally scoped variables in that way, but I'm not super familiar with those cases and since here those variables are passed directly to the `create_map()` function, I would say we should explicitly pass them as arguments (defaults can still be set globally, that's fine, just explicitly pass them in `main:`. In fact, this is discussed in the article that @turnerm linked to [about "tramp" data.](https://softwareengineering.stackexchange.com/questions/335005/is-there-a-name-for-the-anti-pattern-of-passing-parameters-that-will-only-be/335013#335013)

Here's the comment specifically about this issue:

> This is an *improvement* over using global variables. It makes it clear exactly what state each function depends on (and is one step on the road to pure functions). I've heard it called "threading" a parameter through, but I don't know how common this terminology is.

### Calling `create_map()` from `compute_trigger()`

Of course, this all comes back to calling `create_map()` from within `compute_trigger()`. I think ideally, this might be a scenario where really we want an object that can be used to call `create_map()` and `compute_trigger()` separately, without having to separately call their arguments each time. However, given that this is just going to be run twice, not sure it's worth completely refactoring to that.

Since, `gdf_adm` is calculated within `compute_trigger()`, I would argue you should call `create_map()` from within. Otherwise, you could pass the `gdf_adm` to both `compute_trigger()` and `create_map()` and call them separately from `main()`, however, I think that's less ideal until they were class methods.

Tinkaa:
@caldwellst thanks!
I like the caption idea, changed it. 
I am not 100% clear what you mean on the global constants part but I gave them as inputs now. To be honest, I don't fully understand yet when and how to use these global constants and def want to learn more about that. For now since it is kind of time-sensitive leaving it now even though it might be sub-optimal

caldwellst:
Alright @Tinkaa, have reviewed. It seems to be running fine, but there are just a few warnings logged that would be good to confirm we don't need to worry

```shell
>> python src/mwi/compute_trigger_ecmwf.py
/Users/caldwellst/Documents/repos/humdata/pa-anticipatory-action/src/mwi/compute_trigger_ecmwf.py:147: 
UserWarning: Geometry is in a geographic CRS. Results from 'buffer' are likely incorrect. Use 'GeoSeries.to_crs()' to re-project geometries to a projected CRS before this operation.

  area = AreaFromShape(gdf_bound.buffer(3))
INFO:src.indicators.drought.ecmwf_seasonal.ecmwf_seasonal:Downloading ECMWF seasonal forecast v5 for years 2021 - 2022
INFO:src.indicators.drought.ecmwf_seasonal.ecmwf_seasonal:...2021
INFO:src.indicators.drought.ecmwf_seasonal.ecmwf_seasonal:...2022
INFO:src.indicators.drought.ecmwf_seasonal.ecmwf_seasonal:Cannot download data for 2022, because it is in the future. Skipping year 2022.
INFO:__main__:The trigger output has been saved to /Users/caldwellst/Insync/seth.caldwell@humdata.org/Google Drive - Shared drives/Predictive Analytics/CERF Anticipatory Action/General - All AA projects/Data/public/processed/mwi/dry_spells/v1/trigger_metrics/predictive_trigger/mwi_predictive_trigger_20221_unrounded-coords.csv
/Users/caldwellst/Documents/repos/humdata/pa-anticipatory-action/venv/lib/python3.9/site-packages/numpy/lib/nanfunctions.py:1389: RuntimeWarning: All-NaN slice encountered
  result = np.apply_along_axis(_nanquantile_1d, axis, a, q,
INFO:__main__:No data available for 2022-01-01 with leadtime 1 to produce a map. Skipping to next leadtime
INFO:__main__:No data available for 2022-01-01 with leadtime 2 to produce a map. Skipping to next leadtime
INFO:__main__:No data available for 2022-01-01 with leadtime 3 to produce a map. Skipping to next leadtime
```

Primarily these are:

- Buffer warning on geographic CRS.
- All-NaN slice encountered.
- No data available for 2022-01-01. Since it says above that we are skipping 2022, should this still be happening?

You know better if there are any issues with these or if they are harmless, as I haven't really worked with the ECMWF data before. If all good to go, then I think merge. We can give a more thorough review of the code style and everything when refactoring to `aa-toolbox` since for now this is a bit time sensitive.

Tinkaa:
Thanks! Yes the warnings make sense (but could be better handled).
>Buffer warning on geographic CRS.

We are taking a buffer in a degrees CRS while it is meant to be in a meters CRS. It doesn't really matter as we don't want a specifically measured area, just some space around the country boundaries to make sure we include all relevant data
> All-NaN slice encountered.

This is because for the target date (2022-01) there is not data for all leadtimes
> No data available for 2022-01-01. Since it says above that we are skipping 2022, should this still be happening?

Ah yes this is confusing. There is two types of dates, the one that we are attempting to predict for (2022-01) and the date the data was released (which is what skipping 2020 refers to). Changed it to `Cannot download data published in 2022`

caldwellst:
Ah okay great. Yeah, makes sense about the buffer, but could we silence the warning then? If too much a pain, go ahead and merge!

caldwellst (approved):

Tinkaa:
Sorry, leaving it for  now

PR #236, opened 2021-11-08
Feature/ecmwf compare meth

Tinkaa:
Does two things: 
- I recomputed the stats csv's using the new raster function (new is several months old), which changes the column names, so changed that correspondingly in the notebooks
- A while ago I did [an analysis](https://docs.google.com/presentation/d/10IJDJhFPNoo8L4z959VAACarvQMJjUsP3OVBdST41B8/edit#slide=id.gf322d514b3_0_11) on the two methods of downloading data we have used with rounded and unrounded coordinates. I created a notebook to reproduce the figures used in that analysis


caldwellst (approved):
All looks good and is re-running on my machine.

PR #237, opened 2021-11-09
Update MDG analysis with new data to November 8 2021

caldwellst:
New data was received and Leonardo asked me to update the analysis, so I updated the code + slides. No actual change, essentially, the only difference was a couple additional cases at existing peaks in 2017 (they fixed some of the issues with ADM2 codes in the data) and new data (no new cases) until yesterday.


Tinkaa (approved):
Nice

PR #238, opened 2021-11-11
Feature/mwi weighted average

Tinkaa:
First attempt to weighted-averaging the ecmwf forecasts

@turnerm created functionality to compute the weighted average. I also added some functionality to divide the raster cells into smaller units with the `resolution` arg. The weighted average is the same as using `resolution` with the limit going to zero. 

The code is not beautiful, but it seems to work fine. The code with `weighted_average=True` is veeery slow (approx 3 mins per date) so that is gonna take about 12 hours to compute.. 

Also tested for few dates the difference between using `resolution=0.1` and `weighted_average=True` and there is some difference but it is small. 

I computed the stats for all dates with `resolution=0.1` and retrieved the confusion matrices for those, see below. The bad news is that those are even worse than using only the centres of the cells with the "rounded method" (which is already worse than only using the centres of the unrounded method)  

Could you check if the code makes sense and any thoughts on what else to do? 

afbeelding

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ matplotlib==3.4.2
 numpy==1.19.1
 scipy==1.6.1
 rasterio==1.1.7
-xarray==0.16.2
+xarray==0.18.2
 geopandas==0.8.1
 coloredlogs==14.0
 rioxarray==0.8.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ rasterio==1.1.7
 xarray==0.16.2
 geopandas==0.8.1
 coloredlogs==14.0
-rioxarray==0.1.1
+rioxarray==0.8.0
 tqdm==4.50.2
 rasterstats==0.14.0
 cdsapi==0.5.1
--- a/src/indicators/drought/ecmwf_seasonal/processing.py
+++ b/src/indicators/drought/ecmwf_seasonal/processing.py
@@ -1,4 +1,3 @@
-import itertools
 import logging
 import os
 import sys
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -5,7 +5,7 @@ rasterio==1.1.7
 xarray==0.16.2
 geopandas==0.8.1
 coloredlogs==14.0
-rioxarray==0.1.1
+rioxarray==0.8.0
```
turnerm, line 8: You should also update the xarray version

caldwellst:
Hi @Tinkaa, nothing I think too critical in the code that I noticed from scanning through. From the results above, I think it is not worthwhile to switch to this method. Did you originally try using all touching raster cells? Or maybe all touching raster cells where a certain % of the cell is overlapping with the southern region (to avoid where just a tiny sliver/tip is covered)? That's the only other thing I can think of to try that might be useful.

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -195,19 +201,90 @@ def compute_stats_per_admin(
...
+                    )
+                    df_lt_list.append(df_lt)
+                df = pd.concat(df_lt_list)
+                df = df.merge(
+                    gdf_adm[[pcode_col] + add_col], on=pcode_col, how="left"
+                )
+                df["date"] = date_dt
+                df.to_csv(output_path, index=False)
+            else:
+                df = pd.DataFrame()
```
turnerm, line 239: I would remove the weighted average since it's not practical to use.

turnerm (approved):

Tinkaa:
Attempted to clean up as much as possible. Will merge

PR #239, opened 2021-11-15
run all methods

Tinkaa:
- run all variations of methods for the dates we are interested in. Previously I was changing the parameters and running the script, but now it is more transparent which different variations were tested
- make `use_unrounded_area_coords` a private constant instead of global as we are adjusting its value

Could you have a double-check if the logic of running with the different parameters makes sense?


caldwellst:
Hi Tinka, all the logic looks good and re-runs. We can leave all the rest of the cleanup for another time!

PR #241, opened 2021-11-18
TCD hist precip analysis

Tinkaa:
Notebook that processes and analysis historical precipitation, with especially a focus on the occurence of below average precipitation. 
Also compares to a list of historical drought years. 
Serves as a basis to compare precipitation forecasts to.

### DIFF
--- a/analyses/tcd/notebooks/forecasts/tcd_seas_bavg_precip_observed.md
+++ b/analyses/tcd/notebooks/forecasts/tcd_seas_bavg_precip_observed.md
@@ -31,6 +31,7 @@ import matplotlib.pyplot as plt
 import seaborn as sns
 import math
 from dateutil.relativedelta import relativedelta
+import calendar
 
 import altair as alt
 #to plot maps with altair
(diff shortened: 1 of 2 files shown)


caldwellst:
Hey @Tinkaa this looks great! These graphs and text are really comprehensive and found quite easy to understand.

I think my only request at the moment would be to expand the analysis comparing to the list of drought years. Obviously we may or may not be able to trust it, would need to see with the country team, but would want to see at least maybe a basic confusion matrix. I would be interested to see how often we're aligning there. As well, is there any connection with lagged (by year) below average precipitation and drought? Or maybe consecutive poor seasons across multiple years? Maybe is too much for this work at the moment, but think at least the confusion matrix would help illuminate your analysis compared to the impact data we have.

We can use the Biomasse data as impact data after Thursday when it's ready (can turn into yearly data I think), so could make the code flexible to easily change the years or the input data maybe?

Tinkaa:
Thanks! I elaborated a bit by creating the confusion matrices and creating bar plots for only the seasons of interest, which makes it easier to see the consecutive "bad" seasons. Didn't look into lagged yearly precipitation (yet). Is this extra stuff useful? Decided to not go too deep into it for now as I think we have to decide as a team how much attention we want to pay to this piece. 

100% in favor of relating it to Biomasse, nice idea!

caldwellst (approved):
Hi Tinka! So, just some final comments that might be useful before applying this also to BFA. Generally approved, just think these could also be possibly relevant there.

### Precipitation map

As you've noted, could be interesting to look at min/max or other measures that might better show the variance of the start and end of the rainy season. However, beyond that, could be interesting just to look over our area of interest as the scale it's currently mapped out means it's hard to differentiate values in our area of interest which are always quite a bit lower than the high rainfall values further south. Might help us see more clearly patterns in the area of interest or maybe highlight lower levels of rainfall closer to the start or the end of the season?

### Mixing the two periods

When calculating RP, I think does make sense to merge the two forecasting periods since we are going to have the same RP for both. If for any reason we wanted to use separate RPs for the two periods, which I think might not be a good idea for a few reasons, then would make sense to split out.

### Multiple periods to trigger

Re-reading through the analysis, I think my concern is that with this trigger, the rainfall totals are quite variable between the JJA and JAS periods. We should be careful how we design the triggers and their combined use, because based on those confusion matrices, it seems we risk the classic "broken clock right twice a day" scenario. Some things that could be interesting to look at if we have time in this situation:

- Performance where we base a trigger on **both** seasons meeting the below average precipitation.
- Performance, just for the second period, where we base the threshold on the first periods forecast. Could you use a lower threshold? Do we see that if the first period wasn't below average, that basically there is no chance of drought (so we don't even need the second period for instance)? Maybe the relation to the first period isn't based on the trigger threshold but another separate analysis that makes it clearer and easier to set a trigger in the second period?

I think this could eventually be interesting to look into if we have time, as might help us clarify exactly how we expect and recommend these sequential triggers to interact when we know that they are not independent phenomena.

Tinkaa:
You are pro, these are super good points! The best thing I can think of now is to make an issue about it and base further analysis on yor points. Let me know if you think there is a better structure to document these kind of thoughts.

caldwellst:
No, use of a GitHub issue to store these ideas for now seems good!

PR #242, opened 2021-11-19
Feature/tcd explore iri

Tinkaa:
- Analyze historical raster values of IRI forecasts in Chad
- Test different methods of aggregation from raster to region
- Analyze the range of probability values and their statistics after aggregation
- Summarize questions to be discussed by the team / with experts

Ready for review but more from a conceptual level. Some comments should later be removed so keeping it a draft PR for now.


caldwellst:
Hi @Tinkaa, thanks for this! Have reviewed from a more conceptual level as you said above, but still a couple of practical comments. The main one for me would be it would be helpful if you clearly label graphs with titles/axis labels so it's clear what they are showing. For instance, on histogram of the % thresholds, it's not immediately clear that threshold is 40% below average in the forecast, and 20% is referring to geographical areas. It's all eventually clear, but clearer graphs would be great. On that same note, just a bit of text following graphs explaining the results and/or your interpretation would similarly help me easily process what I'm seeing quicker as I scroll through the notebook!

That being said, I think the graphics produced are all very handy, and I like the `hvplot` for plotting the distribution of raster values! However, I think I would be hesitant to use this data at all without additional evidence.

> Is it even worth using the forecasts if the GROC is <=0.5 (=grey in map)?

This was the first thing I also noticed, especially over the areas where we would be operating.

> Can we set the probability threshold based on meteorological knowledge?

I would argue no. While this is important, there are many more factors related to drought (look at the additional data sources we are using), and we are not even touching on the multi-year impact of consecutive dry seasons which is often the driving factor behind the most severe droughts. While meteorological knowledge is super helpful in how we apply this, I think we need to be going beyond. However, that also requires data on drought to validate our methods against.

> What is the preferred aggregation method?

I say impossible to answer without any data to validate our analysis against.

> Should we instead of looking at tercile forecasts look at the [flexible forecasts](https://iridl.ldeo.columbia.edu/maproom/Global/Forecasts/NMME_Seasonal_Forecasts/precip_full.html)?

From looking at the website linked above, it appears that they don't even produce flexible forecasts for Chad except in the very far south outside of our area of focus, because it's "over dry land".

image

> Would IRI be the best seasonal source or are there other sources that might be better?

Maybe this can be answered just by a meteorological perspective, in terms of actual precipitation forecasting skill, but I still think we are missing impact data to triangulate and validate all of our methods.

----

Overall, I think what you've done is great, just that we can't do more with this without impact/observational data for triangulation. And based on what we saw in Malawi, I do not think we should be using forecasts without being able to ensure they are accurate forecasting what we are looking for. This is especially important when the trigger is being designed to become a complex combination of a variety of indicators.

caldwellst:
Maybe a next stage could be the use of the Biomasse data as a triangulation tool for the forecast as well as PRESASS?

Tinkaa:
Thanks! You are 100% right and attempted to make the flow and labelling better

Regarding skill and if we can set any thresholds based on the limited data, I largely agree. Though that would also mean never to use IRI forecasts as we simply don't have more historical data of them, and I think that conclusion is a bit too drastic. But lets have a discussion what we can still do and how! 

Regarding the flexible forecast, it is not produced for Dec-Feb which you selected in the graph as during that period so little rain is generally observed that a dry mask is applied. But if you select the forecast during the rainy season, it does cover Chad. However, not sure if it is of any use to move to the flexible forecasts when we already have so little to validate the tercile forecasts on. 

afbeelding

Tinkaa:
@caldwellst I added a bit of code to compute the lower tercile bounds of observed precipitation and plot them. And to look at the probability distribution of IRI all across the world, but not sure yet if that is useful. 
If you got the time, let me know what you think and if any improvements should be made before merge

caldwellst:
@Tinkaa This looks great! Do we want to adjust this to use the exact area of interest (to be confirmed on Monday, but my understanding of it put in `processed/tcd/cod_ab` in the column of the ADM2 `.gpkg`. I think this could be important because the region of interest cuts out some of the southern areas that might receive most rainfall, so your observed RP events might be changing up.

Also, you could re-calculate the co-occurrence with drought using the observed drought years from Biomasse, just to see how it performs against those. Be good to discuss Monday how we use the list of drought years vs. use something like end of season Biomasse.

Tinkaa:
Yes agree to the adm boundaries and Biomasse! I created an issue for this. Will get to it this week, but just to circumvent never-ending PRs :) Will merge this now and create a new branch for the issue

PR #243, opened 2021-11-23
Add option to get ECMWF API seasonal precip

turnerm:
Expanded `get_ecmwf_forecast` methods to also get the ECMWF API version. Lots of hard-coded stuff which is bad because presumably this will all be migrated soon.

### DIFF
--- a/src/indicators/drought/ecmwf_seasonal/processing.py
+++ b/src/indicators/drought/ecmwf_seasonal/processing.py
@@ -23,13 +23,14 @@ logger = logging.getLogger(__name__)
 
 
 def get_ecmwf_forecast(
-    country_iso3: str, version: int = 5, **kwargs
+    country_iso3: str, version: int = 5, source_cds: bool = True, **kwargs
 ) -> xr.Dataset:
     """
     Retrieve the processed dataset with the forecast for each
     publication date and corresponding lead times
     :param country_iso3: iso3 code of the country of interest
     :param version: version of the ecmwf model to use
+    :param source_cds: whether the data comes from CDS, or the ECMWF API
     :param kwargs: other args that can be given to EcmwfSeasonalForecast()
     :return: dataset with the ecmwf forecasts
     """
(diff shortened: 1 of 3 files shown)


Tinkaa:
This works which is great! And I love how you finally made messy code yourself haha. Lets switch `ecmwf_test` to `ecmwf` and then it is fine for now

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -39,29 +40,44 @@ def get_ecmwf_forecast(
...
-    )
+    if source_cds:
+        ds_ecmwf_forecast = ecmwf_forecast.read_processed_dataset(
+            country_iso3=country_iso3,
+            version=version,
+        )
+    else:
+        dataset_path = (
+            Path(os.environ["AA_DATA_DIR"])
+            / f"private/processed/{country_iso3}/ecmwf_test/"
```
Tinkaa, line 51: change ecmwf_test to ecmwf to make it a bit more ready to use :p

src/indicators/drought/ecmwf_seasonal/processing.py
```diff
@@ -39,29 +40,44 @@ def get_ecmwf_forecast(
...
 
     ds_ecmwf_forecast = get_ecmwf_forecast(
-        country_iso3=country_iso3, version=version, **kwargs
+        country_iso3=country_iso3,
+        version=version,
+        source_cds=source_cds,
+        **kwargs,
     )
     ds_ecmwf_forecast_dict = dates_per_leadtime(ds_ecmwf_forecast)
     return convert_dict_to_da(ds_ecmwf_forecast_dict)
```
Tinkaa, line 83: It is now adding all the `time` steps between the first and last date of forecast that is includes in `ds_ecmwf_forecast`. I would argue it might be nice and clean to remove the empty `time`. Would you agree? 
Can be done by changing this line to `return convert_dict_to_da(ds_ecmwf_forecast_dict).dropna(dim="time",how="all")`

src/indicators/drought/ecmwf_seasonal/processing.py
turnerm: Done!

src/indicators/drought/ecmwf_seasonal/processing.py
turnerm: Good catch, done!

Tinkaa:
@turnerm comparison looks good! 

Also added the `source_cds` arg when computing the stats. Could you have a look if that is fine? After that it is ready to merge imo

Tinkaa (approved):

PR #248, opened 2021-12-06
test realtime data

Tinkaa:
First test to understand the formatting of the seasonal forecast received in realtime from ECMWF.
Plust determine whether the trigger is met for February with the December forecast. 

Would be great if you could have a second opinion whether the processing and the statistic results make sense. 
Of course this needs to be incorporated in our pipeline but that requires quite some work.

### DIFF
--- a/analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
+++ b/analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
@@ -76,8 +76,11 @@ The files are very cryptic but they contain a dataType which can be "em" or "fcm
 Then there also is the "numberOfPoints". This has to do with the geographical area, since we receive the forecast for several geographical areas. 
 Just by testing and looking at the latitude/longitude range I figured 384 is the one matching Malawi but no idea if there is a better way to set this
 
+
+Lastly, cfgrib automatically loads a coordinate names `valid_time` which is then loaded as the date after the end date of the forecast validity, which is very confusing. Due to [this Github issue](https://github.com/ecmwf/cfgrib/issues/97) we discovered you can set backend kwargs on which time dimensions to load. Now `forecastMonth` indicates the leadtime (ranging from 1 to 7) and `verifying_time` indicates the start date of the month the forecast is valid for. 
+
 ```python
-xr.load_dataset(filepath_list[0],engine="cfgrib",filter_by_keys={'numberOfPoints': 384, 'dataType': 'fcmean'})
+xr.load_dataset(filepath_list[0],engine="cfgrib",filter_by_keys={'numberOfPoints': 384, 'dataType': 'fcmean'},backend_kwargs=dict(time_dims=('time', 'forecastMonth','verifying_time')))
 ```
 
 Next we load all files at once
@@ -86,18 +89,10 @@ Next we load all files at once
 def _preprocess_monthly_mean_dataset(ds_date: xr.Dataset):
         
          return (
-            ds_date.assign_coords(
-                {
-                    #get the difference between the valid time month and the publication month 
-                    #valid time is one month ahead of the start of the actual valid time
-                    #so a leadtime of 1 means that the forecast is forecasting the same month as the publication month
-                    "step": (ds_date.valid_time.dt.month-ds_date.time.dt.month)+12*(ds_date.valid_time.dt.year-ds_date.time.dt.year)
-                }
-            )
+            ds_date
             .expand_dims("time")
-            #valid time is one month ahead of the start of the actual valid time --> only confusing so drop
             #surface is empty
-             .drop_vars(["valid_time","surface"])
+             .drop_vars("surface")
         )
 ```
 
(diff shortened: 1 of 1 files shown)


analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
```diff
@@ -0,0 +1,256 @@
...
+```python
+def _preprocess_monthly_mean_dataset(ds_date: xr.Dataset):
+        
+         return (
+            ds_date.assign_coords(
+                {
+                    #get the difference between the valid time month and the publication month 
+                    #valid time is one month ahead of the start of the actual valid time
+                    #so a leadtime of 1 means that the forecast is forecasting the same month as the publication month
+                    "step": (ds_date.valid_time.dt.month-ds_date.time.dt.month)+12*(ds_date.valid_time.dt.year-ds_date.time.dt.year)
```
turnerm, line 94: Here can't you do `(ds_date.valid_time.dt.month-ds_date.time.dt.month -1) % 12` ?

analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
```diff
@@ -0,0 +1,256 @@
...
+            #valid time is one month ahead of the start of the actual valid time --> only confusing so drop
+            #surface is empty
+             .drop_vars(["valid_time","surface"])
+        )
+```
+
+```python
+#i would think setting concat_dim=["time","step"] makes more sense 
+#but get an error "concat_dims has length 2 but the datasets passed are nested in a 1-dimensional structure"
+#it seems to work thought when using concat_dim="time" but would have to test once we have data from several dates..
```
turnerm, line 107: I think this error makes sense because `time` and `step` are a single dimension

analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
turnerm: Oops nevermind, the mod doesn't work and also as you explained, both `time` and `valid_time` are ahead by one month

turnerm (approved):
Looks good, but as we discussed it would be good to confirm with ECMWF what we are interpreting the time coordinates correctly

turnerm:
Confirming that also [here](https://github.com/OCHA-DAP/pa-aa-toolbox/pull/36/files#diff-61f7c6d8e5cc5fde2836a47d4caaac68a2eefae2a1113070a37e483e7ed945eeR222-R237) in the ECMWF API the `time` parameter corresponds to the forecast run date, which is the same as the first forecast date.

I'm wondering if we should move to setting `time` to be the baseline time, i.e. the time you add `step` to in order to get the forecast date. Otherwise `step` doesn't make much sense, unless you know that `time` is actually offset. What do you think?

Tinkaa:
> I'm wondering if we should move to setting time to be the baseline time, i.e. the time you add step to in order to get the forecast date. 

Do you mean with `baseline time` it would be `time - 1 month` ? 
I do think that it would make more sense to have `time+step=forecast_date` but instead of changing `time` I would change `step` . I.e. 0-index instead of 1-index step. I am not sure if we mean the same thing or not?

analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
Tinkaa: I don't think I fully understand. They are both individual dimensions right? Aah or wait is it the difference between coords and dim? 
From this error, do you then understand how we could merge the different files based on both `time` and `step`?

analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
Tinkaa: Due to the github issue solution we found this is no longer relevant. Changed the code to load the times correctly (and not use `valid_time` anymore)

Tinkaa:
>but as we discussed it would be good to confirm with ECMWF what we are interpreting the time coordinates correctly

As we discussed bilaterally we found that the issue is with `cfgrib` which loads `valid_time` differently than what you would expect. This is explained in [this GH issue](https://github.com/ecmwf/cfgrib/issues/97). 
I now changed the code to use `forecastMonth` and `verifying_time` instead of `valid_time` as explained in the linked issue.

analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
turnerm: I was just using dimension in the general sense (not the xarray definition). What I mean is that they are interchangeable, rather than orthogonal dimensions, so you really only need to specify one and the other will be known. I hope that makes sense.

turnerm (approved):
> Do you mean with baseline time it would be time - 1 month ? I do think that it would make more sense to have time+step=forecast_date but instead of changing time I would change step . I.e. 0-index instead of 1-index step. I am not sure if we mean the same thing or not?

I would keep `step` the way it is, as it's supposed to represent the number of months in the future the forecast is for. So `step=1` means the forecast is 1 month in advance, and I think using 0-indexing would just make this confusing. 

This would mean setting the `time` parameter back one month. What do you think? But anyway it's not needed for this PR, perhaps for the pipeline though.

analyses/mwi/dryspells_trigger/exploration/ecmwf_realtime_test.md
Tinkaa: Not sure I understand. Do you mean that individually they are not unique but together they are, so that is automatically determined if one of them is passed?

turnerm: Yes I think so. One of them is just a simple transformation of the other. You could remove one and the dataset would still be fully defined.

Tinkaa: >You could remove one and the dataset would still be fully defined. 

Hmm but I wouldn't agree with that. `time` only contains one value per file, namely the publishing month. While `step` contains 7 values indicating the leadtimes. If we would remove one of the two, we couldn't fully define the dataset.

turnerm: Ah shoot, I'm so sorry Tinka, I was mixing up (what are now) `time` and `verifying_time`. My arguments were all based on `verifying_time`, I hope that makes more sense.

In the case of `time`, I think if you had other files where `time` would be different, it would need to be included in the list of `concat_dim`. However, since there is only one value for `time` then it doesn't treat it as an extra dimension.

Tinkaa: Yes that might be the case

Tinkaa:
>This would mean setting the time parameter back one month.
I wouldn't do that though.. cause then it is confusing to me at least what `time` means and I haven't seen anyone else do it.. You would do it because then `time` represents the month the model was run?

turnerm:
> I wouldn't do that though.. cause then it is confusing to me at least what time means and I haven't seen anyone else do it.. You would do it because then time represents the month the model was run?

I guess my reasoning is that in order to know which month the forecast is referring to, you need to combine `step` with a baseline. Maybe instead of modifying `time` we could add something like `baseline_time`? Although now that I think about it we do have `verifying_time` so perhaps it's not needed.

Tinkaa:
All the processing and plotting has now been implemented in scripts. The processing is done in `pa-aa-toolbox` see [this PR](https://github.com/OCHA-DAP/pa-aa-toolbox/pull/37). The plotting and computing the trigger is done in #253 
I would therefore propose to not merge this notebook to `develop` and thus close the PR. Would you agree?

turnerm:
@Tinkaa yes good to close!

PR #249, opened 2021-12-09
Feature/tcd biomasse wrsi

caldwellst:
Here's the draft code for loading, processing, and comparing Biomasse and WRSI.

### DIFF
--- a/analyses/tcd/notebooks/forecasts/tcd_wrsi.md
+++ b/analyses/tcd/notebooks/forecasts/tcd_wrsi.md
@@ -1,0 +1,202 @@
+This notebook explores the WRSI trigger related to Chad. Given that the WRSI is not a forecast based method (although the extended version does include historical precipitation averages), I have been unable to find any evaluations of its accuracy or skill. We will thus dive straight into understanding how the historical patterns look in Chad and what a reasonable threshold and trigger might look like.
+
+```python
+ #### Load libraries and set global constants
+
+ %load_ext autoreload
+ %autoreload 2
+
+ import geopandas as gpd
+ from shapely.geometry import mapping
+ import pandas as pd
+ import rioxarray
+ import rioxarray.merge
+ import numpy as np
+ import xarray as xr
+ import seaborn as sns
+ import cftime
+ import calendar
+ from dateutil.relativedelta import relativedelta
+ from matplotlib.colors import ListedColormap
+ from rasterio.enums import Resampling
+ import hvplot.xarray
+ import altair as alt
+
+ from pathlib import Path
+ import sys
+ import os
+
+ path_mod = f"{Path(os.path.dirname(os.path.abspath(''))).parents[2]}/"
+ sys.path.append(path_mod)
+ from src.indicators.drought.config import Config
+
+ from src.indicators.drought.wrsi import load_wrsi, filter_wrsi, wrsi_percent_below, _processed_dir
+ from src.indicators.drought.biomasse import load_biomasse_mean
+ from src.utils_general.raster_manipulation import compute_raster_statistics
+
+ hdx_blue="#007ce0"
+
+ iso3="tcd"
+ config=Config()
+ parameters = config.parameters(iso3)
+ country_data_processed_dir = Path(config.DATA_DIR) / config.PUBLIC_DIR / config.PROCESSED_DIR / iso3
+ adm1_bound_path=country_data_processed_dir / config.SHAPEFILE_DIR / "tcd_adm2_area_of_interest.gpkg"
+
+ #### Set variables
+
+_save_processed_path = os.path.join(path_mod, Path(config.DATA_DIR), config.PUBLIC_DIR, config.PROCESSED_DIR, iso3, "wrsi")
+```
+
+Let's load the WRSI data and then plot just the end of season WRSI, to first inspect how it's behaving across the years. We will also subset to the relevant administrative areas. For now, we will just look at croplands since that covers the dominant portion of our area of interest.
+
+```python
+gdf_adm1=gpd.read_file(adm1_bound_path)
+gdf_reg=gdf_adm1[gdf_adm1.area_of_interest == True]
+
+da_crop = load_wrsi("cropland", "current")
+da_crop_clip = da_crop.rio.clip(gdf_reg["geometry"], all_touched=True)
+
+ #not very neat function but does the job for now
+def plt_wrsi(da, subtitle, plt_levels):
+    g = da.plot(
+        col="time",
+        col_wrap=7,
+        levels=plt_levels,
+        cbar_kwargs={
+            "orientation": "horizontal",
+            "shrink": 0.8,
+            "aspect": 40,
+            "pad": 0.1,
+            'ticks': plt_levels,
+        },
+        figsize=(21,10)
+    )
+    
+    for ax in g.axes.flat:
+        gdf_adm1.boundary.plot(linewidth=1, ax=ax, color="grey")
+        gdf_reg.boundary.plot(linewidth=1, ax=ax, color="red")
+        ax.axis("off")
+
+    g.fig.suptitle(f"WRSI published at {subtitle} \n The subtitles indicate the start date",y=1.1);
+
+plt_levels=[0, 60, 100, 250]
+
+plt_wrsi(filter_wrsi(da_crop_clip, 33), "end of dekad 33", plt_levels)
+
+```
+
+```python
+plt_wrsi(filter_wrsi(da_crop_clip, 23), "end of dekad 23", plt_levels)
+```
+
+We can see that there's definitely some spatial patterns in the WRSI, stretching frmo generally satisfied in the southeast to most often not satisfied or even close to satisfied in the northwest. Does this mean we might want to consider only certain areas?
+
+```python
+da_range = load_wrsi("rangeland", "current")
+da_range_clip = da_range.rio.clip(gdf_reg["geometry"], all_touched=True)
+
+plt_wrsi(filter_wrsi(da_range_clip, 33), "end of dekad 33", plt_levels)
+```
+
+Lots of missing data in the NW corner, where the start of season is not clearly occurring by the end of the aseason where the values are calculated. Need to be very careful here, because those areas are essentially always in severe deficit. Should we exclude these from our percentiles and calculations?
+
+```python
+plt_wrsi(da_crop_clip.where(da_crop_clip.time.dt.year == 2021, drop=True), "year 2021", plt_levels)
+```
+
+```python
+plt_wrsi(da_crop_clip.where(da_crop_clip.time.dt.year == 2004, drop=True), "year 2004", plt_levels)
+```
+
+```python
+plt_wrsi(da_range_clip.where(da_range_clip.time.dt.year == 2021, drop=True), "year 2021", plt_levels)
+```
+
+```python
+plt_wrsi(da_range_clip.where(da_range_clip.time.dt.year == 2011, drop=True), "year 2011", plt_levels)
+```
+
+```python
+da_crop_anom = load_wrsi("cropland", "anomaly")
+da_crop_anom_clip = da_crop_anom.rio.clip(gdf_reg["geometry"], all_touched=True)
+
+plt_wrsi(filter_wrsi(da_crop_anom_clip, 33), "end of dekad 33", plt_levels)
+```
+
+```python
+da_range_anom = load_wrsi("rangeland", "anomaly")
+da_range_anom_clip = da_range_anom.rio.clip(gdf_reg["geometry"], all_touched=True)
+
+plt_wrsi(filter_wrsi(da_range_anom_clip, 33), "end of dekad 33", plt_levels)
+```
+
+Let's create a mask based on Biomasse. This will allow us to get rid of areas with no biomasse production and not owrry about the northwest areas that receive extremely limited rainfall.
+
+```python
+bm_mask = rioxarray.open_rasterio(os.path.join(
+    os.getenv("AA_DATA_DIR"),
+    "public",
+    "raw",
+    "general",
+    "biomasse",
+    "BiomassValueMean.tif"
+))
+
+bm_mask = bm_mask.rio.reproject_match(da_range_anom)
+```
+
+```python
+da_range_anom_mask = da_range_anom_clip.where(bm_mask > 0)
+ta = da_range_anom_mask.where(da_range_anom_mask.time.dt.year == 2021, drop = True)
+ta = ta.where(ta.time.dt.month == 11, drop = True)
+ta = ta.where(ta.time.dt.day == 1, drop = True)
+ta.plot()
+```
+
+Great, so we can see above that we get a pretty good cutoff. However, we still have some areas in yellow that are highlighting areas that even by November, the rainy season hadn't started in this year. Let's try to slightly increase the mask.
+
+```python
+da_range_anom_mask = da_range_anom_clip.where(bm_mask > 100)
+ta = da_range_anom_mask.where(da_range_anom_mask.time.dt.year == 2021, drop = True)
+ta = ta.where(ta.time.dt.month == 11, drop = True)
+ta = ta.where(ta.time.dt.day == 1, drop = True)
+ta.plot()
+```
+
+Now with the mask we can see that we're getting very few areas in this dry region, and it might be appropriate to use for our WRSI analysis of rangelands. If we look at croplands:
+
+```python
+da_crop_anom_mask = da_crop_anom_clip.where(bm_mask > 100)
+ta = da_crop_anom_mask.where(da_crop_anom_mask.time.dt.year == 2021, drop = True)
+ta = ta.where(ta.time.dt.month == 11, drop = True)
+ta = ta.where(ta.time.dt.day == 1, drop = True)
+ta.plot()
+```
+
+Here we see almost no reduction in the coverage as there's limited arid regions covering this area. Let's now work on calculating the anomalies for various thresholds in both croplands and rangelands.
+
+```python
+d = wrsi_percent_below(da_range_anom_mask, [70, 80,90,100,110,120])
+d.to_csv(os.path.join(_save_processed_path, "tcd_wrsi_anomaly_rangeland_thresholds.csv"))
+```
+
+Let's do the same for croplands anomaly.
+
+```python
+d = wrsi_percent_below(da_crop_anom_mask, [70, 80,90,100,110,120])
+d.to_csv(os.path.join(_save_processed_path, "tcd_wrsi_anomaly_cropland_thresholds.csv"))
+```
+
+And for the current values.
+
+```python
+# cropland
+da_crop_clip_mask = da_crop_clip.where(bm_mask > 100)
+d = wrsi_percent_below(da_crop_clip_mask, [50, 60, 70, 80,90,100,110,120])
+d.to_csv(os.path.join(_save_processed_path, "tcd_wrsi_current_cropland_thresholds.csv"))
+
+# rangeland
+da_range_clip_mask = da_range_clip.where(bm_mask > 100)
+d = wrsi_percent_below(da_range_clip_mask, [50, 60, 70, 80,90,100,110,120])
+d.to_csv(os.path.join(_save_processed_path, "tcd_wrsi_current_rangeland_thresholds.csv"))
+```
--- a/src/indicators/drought/biomasse.py
+++ b/src/indicators/drought/biomasse.py
@@ -28,6 +28,7 @@ from typing import List, Literal, Union, get_args
 
 import numpy as np
 import pandas as pd
+import rioxarray
 
 from src.utils_general.utils import download_ftp
 
@@ -313,3 +314,12 @@ def load_aggregated_biomasse_data(file_descriptor: str, start_dekad: int = 10):
         admin=file_descriptor, start_dekad=start_dekad
     )
     return pd.read_csv(processed_filepath)
+
+
+def load_biomasse_mean(mask: bool = True):
+    """Used to mask WRSI"""
+    bm_fp = os.path.join(_raw_path, "BiomassValueMean.tif")
+    da = rioxarray.open_rasterio(bm_fp)
+    if mask:
+        da = da.where(da.values <= 0, drop=True)
+    return da
--- a/src/indicators/drought/biomasse.py
+++ b/src/indicators/drought/biomasse.py
@@ -1,0 +1,315 @@
+"""Download and process Biomasse data
+
+The Biomasse data is produced by Action Against Hunger, Spain,
+and made available through the website http://geosahel.info/.
+Documentation of the methodology is made available through
+https://sigsahel.info/wp-content/uploads/2020/12/ACF_BioHydroGenerator_User_Guide.pdf.
+The Biomasse analysis uses Copernicus Global Land Service dry matter
+production (DMP) data: https://land.copernicus.eu/global/products/dmp.
+It is a dekadal product and is updated around 1 week to
+10 days after the end of the period.
+
+While the calculated cumulative biomasse and anomaly data
+is presented at a yearly level on the GeoSahel portal, it
+is not available in dekadal form. The only dekadal form data
+is for DMP values downsampled to the ADMIN0-2 levels across
+the region of interested. The code below accesses this dekadal
+data and then recalculates cumulative biomasse and anomaly
+using that. Since DMP measurements are productivity values, the
+calculation of total biomasse in a year is taken by averaging
+the DMP over the entire period and multiplying by 365.25.
+The default year starts in the 10th dekad (1st of April) until
+the last dekad of March, to correspond with the West African
+monsoon season.
+"""
+
+import os
+from typing import List, Literal, Union, get_args
+
+import numpy as np
+import pandas as pd
+
+from src.utils_general.utils import download_ftp
+
+_base_url = (
+    "http://80.69.76.253:8080/geoserver"
+    "/Biomass/wfs?&REQUEST="
+    "GetFeature&SERVICE=wfs&VERSION=1.1.0"
+    "&TYPENAME=WA_DMP_{admin}_ef_v0&"
+    "outputformat=CSV&srsName=EPSG:4326"
+)
+
+_raw_filename = "WA_DMP_{admin}_ef_v0.csv"
+
+_raw_path = os.path.join(
+    os.getenv("AA_DATA_DIR"), "public", "raw", "general", "biomasse"
+)
+
+
+def _get_raw_path(admin):
+    return os.path.join(_raw_path, _raw_filename.format(admin=admin))
+
+
+_processed_filename = "biomasse_{admin}_dekad_{start_dekad}.csv"
+
+_processed_path = os.path.join(
+    os.getenv("AA_DATA_DIR"), "public", "processed", "general", "biomasse"
+)
+
+
+def _get_processed_path(admin, start_dekad):
+    return os.path.join(
+        _processed_path,
+        _processed_filename.format(admin=admin, start_dekad=start_dekad),
+    )
+
+
+AdminArgument = Literal["ADM0", "ADM1", "ADM2"]
+_valid_admin = get_args(AdminArgument)
+
+
+def _check_admin(admin):
+    if admin not in _valid_admin:
+        raise ValueError("`admin` must be one of 'ADM0', 'ADM1', or 'ADM2'.")
+
+
+def download_dmp(admin: AdminArgument = "ADM2"):
+    """Download raw DMP data
+
+    Raw DMP data is downloaded for specified
+    administrative area. The data is downloaded
+    in CSV format and can be processed using
+    ``process_dmp()``.
+    """
+    _check_admin(admin)
+    url = _base_url.format(admin=admin)
+    raw_path = _get_raw_path(admin)
+    download_ftp(url=url, save_path=raw_path)
+    return None
+
+
+def load_dmp(admin: AdminArgument = "ADM2", redownload: bool = False):
+    _check_admin(admin)
+    raw_path = _get_raw_path(admin)
+    if not os.path.exists(raw_path):
+        download_dmp(admin)
+    elif redownload:
+        os.remove(raw_path)
+        download_dmp(admin)
+
+    df = pd.read_csv(raw_path, na_values=["-9998.8"])
+    df.dropna(axis="columns", how="all", inplace=True)
+    return df
+
+
+def process_dmp(
+    admin: AdminArgument = "ADM2",
+    redownload: bool = False,
+    start_dekad: int = 10,
+):
+    """Process DMP raw data
+
+    DMP raw data is received in a wide format
+    dataset that needs processing. This pivots
+    the data from wide to long format with
+    DMP dekadal observations for all years, while
+    also pivoting the mean values to allow for
+    calculation of biomasse anomaly.
+    """
+    df = load_dmp(admin, redownload)
+
+    # process mean and DMP separately since mean values
+    # are dekadal and DMP are year/dekadal
+    # keep ID column for working with multipolygon admin areas
+    df_mean = df.filter(regex="(^admin|^DMP_MEA|^ID)")
+    df_dmp = df.filter(regex="(^admin|^DMP_[0-9]+|^ID)")
+    admin_cols = [col for col in df_mean.columns if col.startswith("admin")]
+
+    # groupby to average out for the few cases where admin areas
+    # are not contiguous
+    df_mean_long = (
+        pd.wide_to_long(
+            df_mean,
+            i=admin_cols + ["IDBIOHYDRO"],
+            j="dekad",
+            stubnames="DMP_MEA",
+            sep="_",
+        )
+        .reset_index()
+        .drop(labels="IDBIOHYDRO", axis=1)
+        .groupby(admin_cols + ["dekad"])
+        .mean()
+        .reset_index()
+    )
+
+    # calculate anomaly for mean separate from
+    # observed to ensure unique dekadal values
+    df_mean_long["season_index"] = np.where(
+        df_mean_long["dekad"] >= start_dekad,
+        df_mean_long["dekad"] - start_dekad,
+        df_mean_long["dekad"] + (36 - start_dekad),
+    )
+    df_mean_long.sort_values(by=admin_cols + ["season_index"], inplace=True)
+    df_mean_long["biomasse_mean"] = df_mean_long.groupby(by=admin_cols)[
+        "DMP_MEA"
+    ].apply(lambda x: x.cumsum() * 365.25 / 36)
+
+    df_dmp_long = (
+        pd.wide_to_long(
+            df_dmp,
+            i=admin_cols + ["IDBIOHYDRO"],
+            j="time",
+            stubnames="DMP",
+            sep="_",
+        )
+        .reset_index()
+        .drop(labels="IDBIOHYDRO", axis=1)
+        .groupby(admin_cols + ["time"])
+        .sum()
+        .reset_index()
+    )
+
+    # convert time into year and dekad
+    df_dmp_long[["year", "dekad"]] = (
+        df_dmp_long["time"].apply(str).str.extract("([0-9]{4})([0-9]{2})")
+    )
+    df_dmp_long = df_dmp_long.astype({"year": int, "dekad": int})
+
+    # drop values from 1999 that are not from a complete season
+    df_dmp_long = df_dmp_long[
+        ~((df_dmp_long["year"] == 1999) & (df_dmp_long["dekad"] < start_dekad))
+    ]
+
+    # join mean and observed values
+    df_merged = pd.merge(
+        left=df_dmp_long, right=df_mean_long, on=admin_cols + ["dekad"]
+    )
+
+    # calculate biomasse (cumulative sum of DMP across the season
+    # starting with the start_dekad and ending at start_dekad - 1
+    df_merged.sort_values(
+        by=admin_cols + ["year", "dekad"], inplace=True, ignore_index=True
+    )
+    df_merged["season"] = df_merged["year"] + (
+        df_merged["dekad"] >= start_dekad
+    )
+    df_merged["biomasse"] = df_merged.groupby(by=admin_cols + ["season"])[
+        "DMP"
+    ].apply(lambda x: x.cumsum() * 365.25 / 36)
+    df_merged["biomasse_anomaly"] = (
+        100 * df_merged["biomasse"] / df_merged["biomasse_mean"]
+    )
+
+    # re-arrange columns
+    df_merged = df_merged[
+        admin_cols
+        + [
+            "year",
+            "dekad",
+            "DMP_MEA",
+            "DMP",
+            "biomasse_mean",
+            "biomasse",
+            "biomasse_anomaly",
+        ]
+    ]
+
+    # save file to processed filepath
+    processed_path = _get_processed_path(admin, start_dekad)
+    df_merged.to_csv(processed_path, index=False)
+
+    return df_merged
+
+
+def load_biomasse_data(
+    admin: AdminArgument = "ADM2",
+    start_dekad: int = 10,
+    reprocess: bool = False,
+    redownload: bool = False,
+):
+    """Load biomasse processed data
+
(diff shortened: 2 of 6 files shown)


src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+
+While the calculated cumulative biomasse and anomaly data
+is presented at a yearly level on the GeoSahel portal, it
+is not available in dekadal form. The only dekadal form data
+is for DMP values downsampled to the ADMIN0-2 levels across
+the region of interested. The code below accesses this dekadal
+data and then recalculates cumulative biomasse and anomaly
+using that. Since DMP measurements are productivity values, the
+calculation of total biomasse in a year is taken by averaging
+the DMP over the entire period and multiplying by 365.25.
```
Tinkaa, line 20: Aah this took me a while to get. But is it correct that the dekadal DMP is the mean daily production during that dekad? I first interpreted it as the total production during that dekad. But if the first is correct then the calculation of the total biomasse makes sense to me. If so, maybe you could clarify that here?

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+        )
+        .reset_index()
+        .drop(labels="IDBIOHYDRO", axis=1)
+        .groupby(admin_cols + ["dekad"])
+        .mean()
+        .reset_index()
+    )
+
+    # calculate anomaly for mean separate from
+    # observed to ensure unique dekadal values
```
Tinkaa, line 152: this comment is vague to me. Season index is basically a shifted dekad number such that 1 corresponds to the start of the season right?

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    df_merged = pd.merge(
+        left=df_dmp_long, right=df_mean_long, on=admin_cols + ["dekad"]
+    )
+
+    # calculate biomasse (cumulative sum of DMP across the season
+    # starting with the start_dekad and ending at start_dekad - 1
+    df_merged.sort_values(
+        by=admin_cols + ["year", "dekad"], inplace=True, ignore_index=True
+    )
+    df_merged["season"] = df_merged["year"] + (
```
Tinkaa, line 199: I am a bit confused by the definition of season, I would expect it to be one lower.. Cause now e.g. the 2022 season starts in the `start_dekad` of 2021, so on 1 April 2021. It is not super bad but would at least add a comment explaining that. And be careful to later on not mix up season definition

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+        pd.wide_to_long(
+            df_dmp,
+            i=admin_cols + ["IDBIOHYDRO"],
+            j="time",
+            stubnames="DMP",
+            sep="_",
+        )
+        .reset_index()
+        .drop(labels="IDBIOHYDRO", axis=1)
+        .groupby(admin_cols + ["time"])
```
Tinkaa, line 173: you might want to check here first that there are no duplicated `[admin_cols]+["time"]` values? Since we saw this occur once? And now you are summing the values in case there is a duplicate, which I am doubtful whether that is what we should do. (especially since you are taking the mean after the groupby on `df_mean`

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    )
+    df_merged["season"] = df_merged["year"] + (
+        df_merged["dekad"] >= start_dekad
+    )
+    df_merged["biomasse"] = df_merged.groupby(by=admin_cols + ["season"])[
+        "DMP"
+    ].apply(lambda x: x.cumsum() * 365.25 / 36)
+    df_merged["biomasse_anomaly"] = (
+        100 * df_merged["biomasse"] / df_merged["biomasse_mean"]
+    )
```
Tinkaa, line 207: All the calculations look good to me :)

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+def _get_raw_path(admin):
+    return os.path.join(_raw_path, _raw_filename.format(admin=admin))
+
+
+_processed_filename = "biomasse_{iso3}_{admin}_dekad_{start_dekad}.csv"
+
+
+def _get_processed_path(admin, start_dekad, iso3: Union[None, str] = None):
```
Tinkaa, line 57: Commented somewhere else, but could also use `glb` instead of `general`

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    of administrative areas. This should be done
+    by summing up the mean biomasse for each dekad
+    for those areas and the reported biomasse for
+    each year and dekad, then recalculating the
+    anomaly.
+    """
+    _check_admin(admin)
+    df = load_biomasse_data(admin=admin, start_dekad=start_dekad)
+    admin_col = admin.replace("ADM", "admin") + "Pcod"
+    df_subset = df[df[admin_col].isin(admin_pcodes)]
```
Tinkaa, line 289: maybe add a comment here on descripttion that the exact boundaries of the admin shapes that Biomasse uses are not known? So solely concat based on pcode?

Tinkaa: In practice shouldn't give a large difference but think good to note somewhere as a caveat

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    # just keep to year 2000 for unique value per dekad
+    # when calculating biomasse mean aggregated
```
Tinkaa, line 291: maybe add in comment that you select just one year (=2000) since the mean in all years is the same

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    )
+    df_merged = pd.merge(df_dekad, df_mean, on=["dekad"])
+    df_merged["biomasse_anomaly"] = (
+        100 * df_merged["biomasse"] / df_merged["biomasse_mean"]
+    )
+
+    # sorting data again year then dekad
+    df_merged.sort_values(by=["year", "dekad"], inplace=True)
+
+    if iso3 is not None:
```
Tinkaa, line 313: why would you only save it if you have an `iso3`? Maybe in this case do not allow the `iso3` to be `None`?

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+def load_aggregated_biomasse_data(file_descriptor: str, start_dekad: int = 10):
+    processed_filepath = _get_processed_path(
+        admin=file_descriptor, start_dekad=start_dekad
+    )
+    return pd.read_csv(processed_filepath)
+
+
+def load_biomasse_mean(mask: bool = True):
+    """Used to mask WRSI"""
+    bm_fp = os.path.join(_raw_path, "BiomassValueMean.tif")
```
Tinkaa, line 331: where do you get this file from?

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    da = rioxarray.open_rasterio(bm_fp)
+    if mask:
+        da = da.where(da.values <= 0, drop=True)
```
Tinkaa, line 334: I get an error when running `load_biomasse_mean()`. Seems to be caused by the `drop=True`. Assuming you didn't finalize this function yet but fyi

Tinkaa (commented):
I did a thorough review of `biomasse.py` and while I left some comments all computations look good! Let me know if it is ready to not be draft anymore/other files need a draft review.

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+import os
+from typing import List, Literal, Union, get_args
+
+import numpy as np
+import pandas as pd
+import rioxarray
+
+from src.utils_general.utils import download_ftp
+
+_base_url = (
```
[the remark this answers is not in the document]
caldwellst: Will put the global constants in caps, very good point! I put the `_` in front to signal they shouldn't be exported (I think in python modules if the underscore is in front, then doing `import module`, you can never access them using `module._constant`, you can only access explicitly using `from module import _constant`. Do you know if that's right @Tinkaa @turnerm?

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    "/Biomass/wfs?&REQUEST="
+    "GetFeature&SERVICE=wfs&VERSION=1.1.0"
+    "&TYPENAME=WA_DMP_{admin}_ef_v0&"
+    "outputformat=CSV&srsName=EPSG:4326"
+)
+
+_raw_filename = "WA_DMP_{admin}_ef_v0.csv"
+
+_raw_path = os.path.join(
+    os.getenv("AA_DATA_DIR"), "public", "raw", "general", "biomasse"
```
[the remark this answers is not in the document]
caldwellst: It does make sense @Tinkaa, I just totally missed that this existed cause haven't used it previously.

src/indicators/drought/biomasse.py
caldwellst: In this case, we are concating based on Pcode because it's the easiest way to join, but they are the same geographic areas as area of interest. There are some areas outside our area of interest that don't match at the ADMIN2 level but the ADMIN2's completely align for our area of interest in this case.

src/indicators/drought/biomasse.py
caldwellst: Good point!

src/indicators/drought/biomasse.py
caldwellst: In this case @Tinkaa, my thought process was you might want to use the function without saving automatically to a specific country file. Maybe if you were exploring for instance cross-country analysis? Then you would need to manually save because we don't have an immediately logical filename and folder in our structure for it to fit into. What do you think? Previously I actually had it as a filepath argument, but realized in general we are wanting to save to an `iso3` folder.

src/indicators/drought/biomasse.py
caldwellst: Yes good catch!

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,335 @@
...
+    df = load_dmp(admin, redownload)
+
+    # process mean and DMP separately since mean values
+    # are dekadal and DMP are year/dekadal
+    # keep ID column for working with multipolygon admin areas
+    df_mean = df.filter(regex="(^admin|^DMP_MEA|^ID)")
+    df_dmp = df.filter(regex="(^admin|^DMP_[0-9]+|^ID)")
+    admin_cols = [col for col in df_mean.columns if col.startswith("admin")]
+
+    # groupby to average out for the few cases where admin areas
```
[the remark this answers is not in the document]
caldwellst: Happens there and in some other areas in Côte d'Ivoire, have mentioned. It's not a duplicate administrative area, rather that the administrative area is not contiguous so is split into two separate polygons in GIS systems, hence why the output contains multiple rows for them. Isn't a problem at all for us but figured should solve in this manner just in case we eventually use data for those areas.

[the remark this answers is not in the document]
caldwellst: Have amended the comment and code to fit your comments, thanks!

src/indicators/drought/biomasse.py
Tinkaa: Hmm yeah I have been thinking about this as well.. I figured that if the case occurs of cross-country analysis we might as well come up with a fake iso3. So basically that the iso3 arg just serves as an ID. Though that might get a bit ambiguous, especially if we want to check for validity of the iso3 in the toolbox. 

It is just that if we add the None option for iso3 we would basically have to check for that with every dataset that covers more than a country before saving. Which might be a bit ugly. 

However, for now it is fine to leave it in. The thing I would change though then is to make the iso3 a mandatory argument instead of having the default as None. But then keep the option for now to be None instead of a str. Would that make sense? 

@turnerm please chime in if you got any thoughts on this topic beyond this specific case.

src/indicators/drought/biomasse.py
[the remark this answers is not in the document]
Tinkaa: Aaah now I get what you mean, thanks!

src/indicators/drought/biomasse.py
Tinkaa: Aha so this is the same issue with the polygons being split into two. Just make sure that the method after aggregation (I think it should be sum?) is the same with the dmp and mean data

src/indicators/drought/biomasse.py
caldwellst: I like that idea, let's see what @turnerm thinks.

src/indicators/drought/biomasse.py
caldwellst: Yeah, you ar ecorrect, have clarified in the documentation!

src/indicators/drought/biomasse.py
caldwellst: Yeah, I will subtract from 1 so it's clear, but just added it here for the grouping, it wasn't actually included in the output file. However, have decided probably good to add to the output as well, so will make it even clearer by having it be "year1-year2".

src/indicators/drought/biomasse.py
caldwellst: Yes, making sure they are all aligned! We take the mean in the first instance because we want to get a single measure of productivity for each unique area, and summing would be an overstimate. However, now we have a measure of productivity for each area so we want to sum it across our time period.

caldwellst:
I've addressed all your comments @Tinkaa except the ones unresolved, thanks a lot for them! I also went through the NDVI and WRSI files to make those comments also fit there. As well, NDVI processing, downloading, and analysis has been completed and is also part of the PR. Was easier to do all in one PR since most of the work was comparing Biomasse. The rest is ready for review, I just need to go through on Friday and update the documentation of the functions.

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+    processed_path = _get_processed_path(region, type)
+    a = xr.open_dataset(processed_path).to_array("wrsi")
+    a = a.squeeze(drop=True)
+    a.rio.write_crs(_WRSI_CRS, inplace=True)
+    a = a.rio.reproject("EPSG:4326")
+    return a
+
+
+def filter_wrsi(da, dekad):
+    dt = _dekad_to_date(2001, dekad)
```
Tinkaa, line 287: aha so here you select the data from one dekad across all years? Would another option to have an extra coordinate named `dekad` (which is dependent on time) such that you can just filter on dekad. 

If you never use the `time` directly it could also be an option to instead have a `year` and `dekad` dimension but can imagine it is also easy to have them combined into one.

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+    if isinstance(date_obj, str):
+        date_obj = date.fromisoformat(date_obj)
+    year = date_obj.year
+    dekad = (
+        (date_obj.timetuple().tm_mday // 10) + ((date_obj.month - 1) * 3) + 1
+    )
+    return year, dekad
+
+
+def _dekad_to_date(year: int, dekad: int):
```
Tinkaa, line 116: even though private function I would add short docstring if the computed date represents the beginning of the dekad (would assume so) or e.g. the end

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+    filter_month = dt.month
+    filter_day = dt.day
+    return da.where(
+        (da.time.dt.day == filter_day) & (da.time.dt.month == filter_month),
+        drop=True,
+    )
+
+
+def _wrsi_percent_below_thresh(da, threshold):
+    da = xr.where(da <= threshold, 1, 0).mean(dim=["x", "y"])
```
Tinkaa, line 297: nice!

Tinkaa (commented):
Alrighty reviewed `wrsi.py` Seems like many comments but most are small, great work once again! 

Will continue with reviewing the other files

src/indicators/drought/ndvi.py
```diff
@@ -0,0 +1,192 @@
...
+    df = pd.DataFrame(
+        data, columns=["date", "iso3", "anomaly_thresholds", "percent_area"]
+    )
+    df.sort_values(by="date", inplace=True)
+
+    # saving file
+    processed_path.unlink(missing_ok=True)
+    df.to_csv(processed_path)
+
+    return df
```
Tinkaa, line 192: Could you also add a load function?

src/indicators/drought/ndvi.py
```diff
@@ -0,0 +1,192 @@
...
+
+    zf.extract(file_name, _RAW_DIR)
+
+
+def _date_to_dekad(date_obj: Union[date, str]):
+    if isinstance(date_obj, str):
+        date_obj = date.fromisoformat(date_obj)
+    year = date_obj.year
+    dekad = (
+        (date_obj.timetuple().tm_mday // 10) + ((date_obj.month - 1) * 3) + 1
```
Tinkaa, line 77: Maybe I am misinterpreting but why would you use `date_obj.timetuple().tm_mday` instead of `date_obj.day` ? Also applies to the same function in `wrsi.py`

analyses/tcd/notebooks/forecasts/tcd_ndvi.md
```diff
@@ -0,0 +1,85 @@
+```python
```
Tinkaa, line 1: Could you explain a bit what you are doing?

src/indicators/drought/ndvi.py
```diff
@@ -0,0 +1,192 @@
...
+        Path(os.getenv("AA_DATA_DIR"), "public", "processed", iso3, "ndvi")
+    )
+    _processed_dir.mkdir(parents=True, exist_ok=True)
+    return Path(
+        _processed_dir,
+        "{iso3}_ndvi_anomaly.csv".format(iso3=iso3),
+    )
+
+
+def process_ndvi(iso3, geometries, thresholds):
```
Tinkaa, line 167: this does still take quite long to do. Which isn't bad but would add a comment about it. Plus maybe a progress bar could be an idea?

analyses/tcd/notebooks/forecasts/tcd_wrsi.md
```diff
@@ -0,0 +1,212 @@
...
+ta = ta.where(ta.time.dt.month == 11, drop = True)
+ta = ta.where(ta.time.dt.day == 1, drop = True)
+ta.plot()
+```
+
+Here we see almost no reduction in the coverage as there's limited arid regions covering this area. Let's now work on calculating the anomalies for various thresholds in both croplands and rangelands.
+
+```python
+d = wrsi_percent_below(da_range_anom_mask, [70, 80,90,100,110,120])
+d.to_csv(os.path.join(_save_processed_path, "tcd_wrsi_anomaly_rangeland_thresholds.csv"))
```
Tinkaa, line 190: comment out saving paths

analyses/tcd/notebooks/forecasts/tcd_ndvi.md
```diff
@@ -0,0 +1,85 @@
...
+import xarray as xr
+import seaborn as sns
+import cftime
+import calendar
+from dateutil.relativedelta import relativedelta
+from matplotlib.colors import ListedColormap
+from rasterio.enums import Resampling
+import hvplot.xarray
+import altair as alt
+
```
Tinkaa, line 20: There is also lots of libraries that are imported but not used.

analyses/tcd/notebooks/forecasts/tcd_ndvi.md
```diff
@@ -0,0 +1,85 @@
...
+ #### Set variables
+
+_save_processed_path = os.path.join(path_mod, Path(config.DATA_DIR), config.PUBLIC_DIR, config.PROCESSED_DIR, iso3, "wrsi")
+```
+
+```python
+gdf_adm1=gpd.read_file(adm1_bound_path)
+gdf_reg=gdf_adm1[gdf_adm1.area_of_interest == True]
+
```
Tinkaa, line 47: I would also add the function call of `download_ndvi()` and comment it out. But just so the whole process is documented.

analyses/tcd/notebooks/forecasts/tcd_ndvi.md
```diff
@@ -0,0 +1,85 @@
...
+da = process_ndvi("tcd", gdf_reg.geometry, [50, 60, 70, 80, 90, 100])
+
+```
+
+```python
+df1 = rioxarray.open_rasterio(os.path.join(config.DATA_DIR, "public", "raw", "glb", "ndvi", "wa0333pct.tif"))
```
Tinkaa, line 53: What are you doing here and below? Almost seems you are testing the methodology that is now implemented in `process_ndvi`?

analyses/tcd/notebooks/forecasts/tcd_wrsi.md
```diff
@@ -0,0 +1,212 @@
...
+import calendar
+from dateutil.relativedelta import relativedelta
+from matplotlib.colors import ListedColormap
+from rasterio.enums import Resampling
+import hvplot.xarray
+import altair as alt
+
+from pathlib import Path
+import sys
+import os
```
Tinkaa, line 29: remove unused libraries

analyses/tcd/notebooks/forecasts/tcd_wrsi.md
```diff
@@ -0,0 +1,212 @@
+This notebook explores the WRSI trigger related to Chad. Given that the WRSI is not a forecast based method (although the extended version does include historical precipitation averages), I have been unable to find any evaluations of its accuracy or skill. We will thus dive straight into understanding how the historical patterns look in Chad and what a reasonable threshold and trigger might look like.
+
```
Tinkaa, line 2: could you add that further exploration is done in the R file?

analyses/tcd/notebooks/forecasts/tcd_biomasse.md
```diff
@@ -0,0 +1,40 @@
...
+adm2_bound_path=country_data_processed_dir / config.SHAPEFILE_DIR / "tcd_adm2_area_of_interest.gpkg"
+
+ #### Set variables
+
+_save_processed_path = os.path.join(path_mod, Path(config.DATA_DIR), config.PUBLIC_DIR, config.PROCESSED_DIR, iso3, "wrsi")
+
+import src.indicators.drought.biomasse as bm
+```
+
+This is just the code for downloading and procesing the Biomasse data, then aggregating to a specific set of admin codes (our region of interest in Chad.
```
Tinkaa, line 30: would add that further exploration is done in the R file

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,362 @@
...
+    _check_admin(admin)
+    url = _BASE_URL.format(admin=admin)
+    raw_path = _get_raw_path(admin)
+    download_ftp(url=url, save_path=raw_path)
+    return None
+
+
+def load_dmp(admin: AdminArgument = "ADM2", redownload: bool = False):
+    _check_admin(admin)
+    raw_path = _get_raw_path(admin)
```
Tinkaa, line 103: hmm don't you want to check here for file existence instead of redownloading every time you load the dataset?

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,362 @@
...
+    """Process DMP raw data
+
+    DMP raw data is received in a wide format
+    dataset that needs processing. This pivots
+    the data from wide to long format with
+    DMP dekadal observations for all years, while
+    also pivoting the mean values to allow for
+    calculation of biomasse anomaly.
+    """
+    df = load_dmp(admin, redownload)
```
Tinkaa, line 127: I would assume the dataset is already downloaded when calling process. So check for file existence and if doesn't exist raise an error.

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,362 @@
...
+    ``process_dmp()``.
+    """
+    _check_admin(admin)
+    url = _BASE_URL.format(admin=admin)
+    raw_path = _get_raw_path(admin)
+    download_ftp(url=url, save_path=raw_path)
+    return None
+
+
+def load_dmp(admin: AdminArgument = "ADM2", redownload: bool = False):
```
Tinkaa, line 101: the redownload arg is never used..

Tinkaa (commented):
Alrighty reviewed everything except for the R scripts. 

And all the computations look good to me, woohoo. 

Only comments on code style and usability. Some things I commented earlier on the wrsi script I also saw in the ndvi script, so can do a second pass once we discussed those comments.

src/indicators/drought/ndvi.py
```diff
@@ -0,0 +1,192 @@
+"""Code to download and process NDVI data
+
+NDVI data is downloadable and available through the USGS
```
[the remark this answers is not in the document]
caldwellst: Hmm not sure I get why? As in explaining that it's smoothed/slightly processed already?

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+        raise ValueError("`region` must be one of 'cropland' or 'rangeland'.")
+
+    if start_date is None:
+        start_year, start_dekad = 2001, 13
+    elif isinstance(start_date, list):
+        start_year, start_dekad = start_date[0], start_date[1]
+    else:
+        start_year, start_dekad = _date_to_dekad(start_date)
+
+    if isinstance(end_date, list):
```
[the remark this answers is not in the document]
caldwellst: The `end_date` structure is slightly amended just because if it isn't a `list`, then both methods need to call `_date_to_dekad()`, so wanted to not repeat myslf by calling it twice. But maybe still clearer to just match and do some slight repetition? What do you think?

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+        end_year, end_dekad = end_date[0], end_date[1]
+    else:
+        if end_date is None:
+            end_date = date.today()
+        end_year, end_dekad = _date_to_dekad(end_date)
+
+    i_year = start_year
+    i_dekad = min(start_dekad, 33)
```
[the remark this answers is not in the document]
caldwellst: So we could restrict the user passed values ,but they might just want to pass in like `2000-01-01` to `2010-12-31` or something to download all dates. And default `end_date` is just `date.today()` which I think is easiest, so wanted to just limited those so we don't get errors. Have explained more in documentation now.

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+    raw_dir = _get_raw_dir(region)
+    raw_dir.mkdir(parents=True, exist_ok=True)
+
+    if not redownload:
```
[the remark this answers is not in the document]
caldwellst: Okay clear! I saw this once so kept using it but had no other logic than that hahahaah. I will see how the `toolbox` is using it.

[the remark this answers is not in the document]
caldwellst: `clobber` sounds cool as shit hahaha. I will go with that. Also makes it quicker to transfer to `pa-aa-toolbox`.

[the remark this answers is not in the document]
caldwellst: Ugh, you're not wrong, so much smoother. I just got into my head and did this because it was slightly more complex with ARC2 where we storing data within a single file, rather than the multiple files here. Refactoring!!!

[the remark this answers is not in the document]
caldwellst: So much shorter and cleaner. NICE.

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+
+    base_url = _BASE_URL.format(region=url_region)
+    return base_url + url_zip_filename
+
+
+def _download_wrsi(year: int, dekad: int, region: str, raw_dir: Path):
+    try:
+        url = _get_url_path(year=year, dekad=dekad, region=region)
+        resp = urlopen(url)
+    except HTTPError:
```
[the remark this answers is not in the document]
caldwellst: So, I don't actually want to raise an error here, I should actually have as `logger.message`. In the previous instance of the data (pre-CHIRPS), there were certain missing years/dekads that weren't downloaded because they didn't exist, but isn't an actual error, just important to catch and log for the user if they might want to query why some data is missing. I put `logger.error` because I was avoiding changing my setting to show messages by default in console lol...But should probably be message so have changed! Hopefully isn't an issue anymore anyway with the CHIRPS data.

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+            f"No data available for region {region} for "
+            f"dekad {dekad} of {year}, skipping."
+        )
+        return
+
+    zf = ZipFile(BytesIO(resp.read()))
+    for type in ["do", "eo", "er"]:
+        file_name = _BASE_FILENAME.format(year=year, dekad=dekad, type=type)
+        save_path = Path(raw_dir, file_name)
+        save_path.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
caldwellst: No need to overwrite. I think I kept it in because I saw old scripts talking about issues with overwriting, and just mindlessly kept that!

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+
+    ad.to_netcdf(processed_path)
+
+    return ad
+
+
+def load_wrsi(region, type):
+    processed_path = _get_processed_path(region, type)
+    a = xr.open_dataset(processed_path).to_array("wrsi")
+    a = a.squeeze(drop=True)
```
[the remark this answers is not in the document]
caldwellst: Yeah, I think it's nicer to explicitly save as dataset, because `DataArray.to_netcdf()` converts to a dataset on the backend, and wasn't sure if there was a better format to save in instead of `NetCDF` (prefer that to saving just a million band `.tif` for instance). But yes, will open with `rioxarray` and see how it's different! These comments are gold, thanks so much.

src/indicators/drought/wrsi.py
caldwellst: Yeah, I have added as a TODO. The reason it was troublesome before, cause I tried, was beacuse then you have multiple dimensions and then I was using rioxarray and that was problematic. Then i switched it all up, a mess. Something to investigate soon, this great!

src/indicators/drought/wrsi.py
caldwellst: Yes, didn't document this well yet, but going through now!

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,309 @@
...
+_BASE_URL = (
+    "https://edcftp.cr.usgs.gov/project/fews/"
+    "africa/west/dekadal/wrsi-chirps-etos/{region}/"
+)
+
+_URL_ZIP_FILENAME = "w{year:04}{dekad}{region}.zip"
+
+_BASE_FILENAME = "w{year:04}{dekad}{type}.tif"
+
+
```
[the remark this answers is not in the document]
caldwellst: Done!!

src/indicators/drought/ndvi.py
```diff
@@ -0,0 +1,192 @@
...
+        end_year, end_dekad = end_date[0], end_date[1]
+    else:
+        if end_date is None:
+            end_date = date.today()
+        end_year, end_dekad = _date_to_dekad(end_date)
+
+    i_year = start_year
+    i_dekad = start_dekad
+
+    if not redownload:
```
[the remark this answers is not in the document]
caldwellst: So much smoother now.

src/indicators/drought/ndvi.py
```diff
@@ -0,0 +1,192 @@
...
+)
+
+_BASE_FILENAME = "wa{year:02}{dekad:02}pct"
+
+_RAW_DIR = Path(os.getenv("AA_DATA_DIR"), "public", "raw", "glb", "ndvi")
+
+
+def _get_paths(year, dekad):
+    base_file_name = _BASE_FILENAME.format(year=year % 100, dekad=dekad)
+    url_zip_filename = base_file_name + ".zip"
```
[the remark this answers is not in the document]
caldwellst: f-string fiend.

src/indicators/drought/ndvi.py
caldwellst: Yeah agreed.

analyses/tcd/notebooks/forecasts/tcd_ndvi.md
caldwellst: Ha yes, was testing so have removed now!

src/indicators/drought/biomasse.py
caldwellst: Was all a bit of a mixup switching to `unlink`, woops!

caldwellst:
Madness. So many good comments, thanks Tinka! Have hopefully addressed almost all, just a couple left with TODO to be done when merging into `pa-aa-toolbox`. Love this continuous learning and thanks again for all the helpful tips.

src/indicators/drought/biomasse.py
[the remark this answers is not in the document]
Tinkaa: btw I learnt from monica that yes. It should have an underscore and capitals

src/indicators/drought/wrsi.py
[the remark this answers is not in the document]
Tinkaa: Aah I see, missed that. Then yes this is fine. Could maybe with the start_date to first `if list` and then `elif is None` to mimic the structure a bit more, but that is very minor.

src/indicators/drought/ndvi.py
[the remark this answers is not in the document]
Tinkaa: Yeah that is what I meant but very minor. Just from a toolbox perspective that people might expect it is the raw NDVI data, but we would possibly explain a bit more in that case anyway

src/indicators/drought/biomasse.py
Tinkaa: @caldwellst I am not sure if I agree with this. I get the mean and the sum in general but when there is duplicate areas, I don't think it is correct. Maybe I am wrong but want to make sure we are not doing smth wrong. 

For example in Liberia we have the admin2 area with pcode `LR0402`. When I use the processed file (`biomasse_general_ADM2_dekad_10.csv`) and check the anomalies for that pcode, I see that they are always 180 or higher. If I understand the anomaly correctly the average anomaly should be 100 and thus it is not correct that it is always higher than 100. I would think the reason is that for the mean you take the mean over the two noncontigunous polygons while for the dekad-yearly DMP you take the sum. What is your take on this?

src/indicators/drought/biomasse.py
```diff
@@ -0,0 +1,432 @@
...
+        da = da.where(da.values <= 0, drop=False)
+    return da
+
+
+def _get_raw_path(admin):
+    return _RAW_PATH / _RAW_FILENAME.format(admin=admin)
+
+
+def _get_processed_path(
+    admin: AdminArgument, start_dekad: int, iso3: Union[None, str] = None
```
Tinkaa, line 413: No need to change, but for learning: the recommended way is to instead of using Union with None to use `Optional` see [here](https://stackoverflow.com/questions/39429526/how-to-specify-nullable-return-type-with-type-hints)

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,382 @@
...
+Both are published using ESRI:102022. There is overlap between the
+two datasets where they don't agree, even though both are
+measuring the WRSI with respect to millet, due to their
+calculations for rangeland and cropland. Since many
+countries in the Sahel overlap with both datasets, both
+should be considered and we should be careful in learning
+to apply and use both products where applications.
+
+I have also just found the CHIRPS WRSI data on the USGS website.
```
Tinkaa, line 28: Is this still relevant?

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,382 @@
...
+        start_year, start_dekad = _date_to_dekad(start_date)
+
+    if isinstance(end_date, list):
+        end_year, end_dekad = end_date[0], end_date[1]
+    else:
+        if end_date is None:
+            end_date = date.today()
+        end_year, end_dekad = _date_to_dekad(end_date)
+
+    i_year = max(start_year, 2001)
```
Tinkaa, line 136: I would move this check already up to the `start_year,start_dekad` determination and thus add an extra line with `start_year = max(start_year,2001)`

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,382 @@
...
+    i_dekad = max(min(start_dekad, 33), 13)
```
Tinkaa, line 137: same here, `start_dekad=...`

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,382 @@
...
+    raw_dir = _get_raw_dir(region)
+    raw_dir.mkdir(parents=True, exist_ok=True)
+
+    # existing file dates
+    dts = [
+        _date_to_dekad(_fp_date(filename))
+        for filename in raw_dir.glob("*.tif")
+    ]
+    while not (
+        i_year > end_year or (i_year == end_year and i_dekad > i_dekad)
```
Tinkaa, line 147: Sorry I thought I commented this before but apparently I didn't. I would recommend using a for loop with a zip over the years and dekads instead of the `while not`. So smth like `for year,dekad in zip(range(min_year,max_year),range(13,33):`

Only for the start_year you then would have to have one check like `if year==start_year and dekad<start_dekad:continue` I would say that is cleaner and prevents you from doing the plus one and checking if dekad is larger than 33 at the end. Does that make sense to you as well?

Tinkaa: Also applies to the ndvi `while not` code

src/indicators/drought/wrsi.py
```diff
@@ -0,0 +1,382 @@
...
+- WA: the largest region, covering from the coast at
+    5 degrees to around 15 degrees latitude, ranging
+    from the western coast of Senegal to the western
+    border of Sudan. This is specifically for croplands:
+    https://earlywarning.usgs.gov/fews/product/56
+- W1: smaller, ranges from around 11 degrees latitude
+    to around 17, and from the eastern border of Chad
+    to Senegal. This is specifically for rangelands:
+    https://earlywarning.usgs.gov/fews/product/57
+
```
Tinkaa, line 19: You mention it later but would also add here that data is only available from dekad 13 till 33 (add dates as well)

Tinkaa (commented):
Super super great work! Added a few more comments to the py files (sorry!) that I only catched now. But think then it is ready to go

Two general comments: 
- would be good to clean-up old files on the GDrive. Saw for example the `general` folder still existing
- When I was playing a little with the Biomasse data, I saw some negative anomaly values. Which were cause by negative Biomasse (mean) values so would be good to double-check that and see what that means.

src/indicators/drought/biomasse.py
caldwellst: Aha yes I see now. I think moving to a weighted average for the DMP values themselves makes sense then. I will run this and test if it's comparable. Thanks for the catch here!

src/indicators/drought/biomasse.py
caldwellst: Brilliant, thank you!!

caldwellst:
Thanks for all the comments! Have resolved them all and hope they are good to go now. Regarding some of your above comments.

- I removed `general` from my machine before leaving, and still don't see it here. Not sure why still on drive, but have removed, hopefully that fully cleans it up.

- Good catch on negative values. I thought the missing values were all -9998.8 but there were also -9999.0 which I wasn't catching. All seems good now when checking, and have used averaging across both the `DMP_MEA` and `DMP` values. Now the checks seem to work out.

I have messaged back to the Biomasse team at ACF so we can discuss with them the methods applied above and confirm that we aren't doing anything wrong.

Tinkaa:
You are amazing, thanks! Last thing I did was try to run the R scripts and only two minor changes to make them run, so committed them. 

One last thing is that we discussed before that it would be nice to compare our own computed biomasse anomaly to the one produced by the platform. Do you agree that that would still be a good sanity check? If so, I think it makes sense to do that in a separate PR and thus merge this one 🎉

caldwellst:
Yes, good idea! Will need to think through how to do that and then try and get that done. Will merge this then do that.

Issue #245, opened 2021-11-30
TCD: analysis precipiation patterns and dependence JJA and JAS


Tinkaa:
Discussion that came out of #241 

For TCD we are looking at below average precipitaiton during JJA and JAS. But 1) we should investigate further if below average is the appropiate measure and 2) we now treat JJA and JAS as independent but they are not. 

### Precipitation map

As you've noted, could be interesting to look at min/max or other measures that might better show the variance of the start and end of the rainy season. However, beyond that, could be interesting just to look over our area of interest as the scale it's currently mapped out means it's hard to differentiate values in our area of interest which are always quite a bit lower than the high rainfall values further south. Might help us see more clearly patterns in the area of interest or maybe highlight lower levels of rainfall closer to the start or the end of the season?

### Mixing the two periods

When calculating RP, I think does make sense to merge the two forecasting periods since we are going to have the same RP for both. If for any reason we wanted to use separate RPs for the two periods, which I think might not be a good idea for a few reasons, then would make sense to split out.

### Multiple periods to trigger

Re-reading through the analysis, I think my concern is that with this trigger, the rainfall totals are quite variable between the JJA and JAS periods. We should be careful how we design the triggers and their combined use, because based on those confusion matrices, it seems we risk the classic "broken clock right twice a day" scenario. Some things that could be interesting to look at if we have time in this situation:

- Performance where we base a trigger on both seasons meeting the below average precipitation.
- Performance, just for the second period, where we base the threshold on the first periods forecast. Could you use a lower threshold? Do we see that if the first period wasn't below average, that basically there is no chance of drought (so we don't even need the second period for instance)? Maybe the relation to the first period isn't based on the trigger threshold but another separate analysis that makes it clearer and easier to set a trigger in the second period?

I think this could eventually be interesting to look into if we have time, as might help us clarify exactly how we expect and recommend these sequential triggers to interact when we know that they are not independent phenomena.

PR #250, opened 2021-12-13
Tercile check notebook

turnerm:
Making a PR for the analysis used to address #245

### DIFF
--- a/analyses/tcd/notebooks/forecasts/tcd_tercile_check.md
+++ b/analyses/tcd/notebooks/forecasts/tcd_tercile_check.md
@@ -1,0 +1,153 @@
+```python
+import xarray as xr
+import matplotlib.pyplot as plt
+import numpy as np
+```
+
+```python
+# Read in monthly Chad dataset
+filename = '/home/turnerm/sync/aa_repo_data/Data/public/processed/tcd/chirps/monthly/tcd_chirps_monthly.nc'
+ds = xr.load_dataset(filename)
+```
+
+```python
+# Simplify into DA at a single datapoint
+da = ds['precip'].isel(latitude=140, longitude=150)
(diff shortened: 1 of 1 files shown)


Tinkaa:
Thank you so much for all of this, learnt a lot! 

Some filepaths were hardcoded so changed them

But more importantly I was playing around with the code and think I found a slightly more efficient way to compute than the already much better method you proposed. Also implemented all the three terciles.

Could you have a look to see if this makes sense? If so, I can change the methodology in the `.py` script. 

One discussion question: say we look at the below normal tercile, then we now keep the original values of all cells in the below normal tercile and set all others to NaN. I was thinking it might make more sense to instead of keeping the original values set them all to 1. Cause then you can just filter the original `precip` var where the `below_normal` var is 1. Would you have a pref for one of the two?

turnerm:
> But more importantly I was playing around with the code and think I found a slightly more efficient way to compute than the already much better method you proposed. Also implemented all the three terciles.
> Could you have a look to see if this makes sense? If so, I can change the methodology in the .py script.

Yes that looks great!

> One discussion question: say we look at the below normal tercile, then we now keep the original values of all cells in the below normal tercile and set all others to NaN. I was thinking it might make more sense to instead of keeping the original values set them all to 1. Cause then you can just filter the original precip var where the below_normal var is 1. Would you have a pref for one of the two?

Do you mean to turn the tercile datasets into booleans and just use those every time you need the actual values? That sounds good to me!

Tinkaa:
>Do you mean to turn the tercile datasets into booleans and just use those every time you need the actual values?

Yes exactly

I will merge and create a new PR to implement it in the code

Tinkaa (approved):

PR #253, opened 2021-12-16
Feature/mwi pred trigger rt

Tinkaa:
Okay I hope this is finally the last PR for the mwi pred trigger.. 

Changed the trigger script to also accept the `source_cds` parameter and thus also be able to compute the trigger values on the high resolution ecmwf data

### DIFF
--- a/src/mwi/get_ecmwf_seasonal_data.py
+++ b/src/mwi/get_ecmwf_seasonal_data.py
@@ -47,7 +47,7 @@ def main(download=False, compute_stats=True, use_cache=True):
         use_unrounded_area_coords=USE_UNROUNDED_AREA_COORDS
     )
     df_country_boundaries = gpd.read_file(ADM0_BOUND_PATH)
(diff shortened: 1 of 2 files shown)


src/mwi/get_ecmwf_seasonal_data.py
```diff
@@ -47,7 +47,7 @@ def main(download=False, compute_stats=True, use_cache=True):
         use_unrounded_area_coords=USE_UNROUNDED_AREA_COORDS
     )
     df_country_boundaries = gpd.read_file(ADM0_BOUND_PATH)
-    if download:
+    if download and SOURCE_CDS:
```
turnerm, line 50: This might be confusing because if the user passes `download=True` they will not know why the download isn't working. Maybe provide an explanation message? (Or at least a comment)

turnerm (approved):
I didn't run it but it looks good!

Issue #246, opened 2021-11-30
Ensure correct computation observed terciles


Tinkaa:
We are using CHIRPS data to compute historically observed below average precipitation. 
This code was written some time ago, but never properly code reviewed. 
Since we are now using it more often, it would be good to have a second pair of eyes looking at it. 

The code can be found [here](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/13f2f3da764b108df659ec3eb9e102a56ad60a5d/src/indicators/drought/chirps_rainfallobservations.py#L195) (line 195-258). 
The output is for example used for Chad [here](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/f4aa628bc185dac6c3d39e388c621ce1ac89500f/analyses/tcd/notebooks/forecasts/tcd_seas_bavg_precip_observed.md)

In general it looks decent to me but why I want to double check is because 
1) in both bfa and tcd we see much more prevalence of below average in the 80's and 90's than in the last 20 years. Maybe precipitation patterns or technologies changed but want to be sure it is not us
2) We don't see great correspondence between years with bavg precipitation and reported drought years. I am not perse surprised about this, but better be sure before we make claims on that. 

As sanity checks I did simple things like making sure the occurence of below average is 1/3. I also manually compared a subset of the dates to the anomalies on [CHC's early warning explorer](https://chc-ewx2.chc.ucsb.edu/) but mainly for more recent dates.

Tinkaa:
![afbeelding](https://user-images.githubusercontent.com/12556634/144046229-7f37517f-6dd8-4da2-95cd-50d7bac1be8e.png)
Actually the prevalance of bavg precipiation in the 80s might be correct, see e.g. this messy graph from [here](https://pubs.usgs.gov/fs/2012/3070/FS2012-3070.pdf)

turnerm:
I've looked very carefully at [this method](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/13f2f3da764b108df659ec3eb9e102a56ad60a5d/src/indicators/drought/chirps_rainfallobservations.py#L195) (worked through an example myself) and it all looks good to me. Some minor comments but I don't think they will have any effect on the results:
- for the quantile function:
  - it could be good to set `skipna = True`
  - instead of passing `0.33` I would pass `1/3` (more precise)
- when you set null values to -666, could that potentially be biasing the sample low? Should these be kept as N/A instead?

I'll check the notebook next

[...]

PR #254, opened 2021-12-20
Feature/chirps tercile

Tinkaa:
Implemented a neater method to compute the terciles as discussed in #246 and #250 

While before we only computed the lower tercile, this code implements the computation of all three terciles. 

Didn't make it super neat code as that has to be reformatted soon for `aa-toolbox` anyway. So more about better method of computation


src/indicators/drought/chirps_rainfallobservations.py
```diff
@@ -214,63 +239,94 @@ def compute_seasonal_lowertercile_raster(
...
 
-    Path(chirps_seasonal_lowertercile_country_filepath.parent).mkdir(
+    Path(chirps_seasonal_tercile_filepath.parent).mkdir(
         parents=True, exist_ok=True
     )
     logger.debug("Computing lower tercile values...")
+    # for some unknown reason the rolling sum takes very long to
+    # compute when using rioxarray so sticking to xarray for now
     ds = xr.load_dataset(chirps_monthly_country_filepath)
+    da = ds.precip
```
turnerm, line 260: You can just tack this on to the end of the previous line and avoid defining `ds`

src/indicators/drought/chirps_rainfallobservations.py
```diff
@@ -214,63 +239,94 @@ def compute_seasonal_lowertercile_raster(
...
+
+    # save tercile boundaries
+    da_season_climate_quantile.to_netcdf(
+        chirps_seasonal_tercile_bounds_filepath
+    )
+
+    da_bn = xr.where(
+        da_season
+        <= da_season_climate_quantile.sel(
+            quantile=1 / 3, month=da_season.time.dt.month
```
turnerm, line 291: Is there a way to rename the quantile variables? It's just that selecting on a float value seems like it's asking for trouble

turnerm (approved):
Looks great! You don't have to do this now, but while it's fresh in your mind, it could be a good moment to add a unit test or two, around the tercile computation logic. It might be worth breaking down the function a bit more to make the testing easier.

src/indicators/drought/chirps_rainfallobservations.py
Tinkaa: Great idea, renamed them to `lower_bound` and `upper_bound`

Tinkaa:
Alrighty that was a long rabbit hole but I created some tests. Along the way split up the functions in the main file. Could you have another review of this? :)

src/indicators/drought/chirps_rainfallobservations.py
```diff
@@ -189,88 +195,153 @@ def get_filepath_seasonal_lowertercile_raster(
...
+        4: "FMA",
+        5: "MAM",
+        6: "AMJ",
+        7: "MJJ",
+        8: "JJA",
+        9: "JAS",
+        10: "ASO",
+        11: "SON",
+        12: "OND",
+    }
```
turnerm, line 250: This may be overkill, but I can't resist automating these things ;)
```python
from calendar import month_name
end_month_season_mapping = { i:
month_name[(i - 2) % 12][0]+ month_name[(i -1) % 12][0] + month_name[i] for i in range(1, 13)
}

Comment (turnerm) [inline src/indicators/drought/chirps_rainfallobservations.py:250]: This may be overkill, but I can't resist automating these things ;)
```python
from calendar import month_name
end_month_season_mapping = { i:
month_name[(i - 2) % 12][0]+ month_name[(i -1) % 12][0] + month_name[i] for i in range(1, 13)
}

File: src/indicators/drought/chirps_rainfallobservations.py [hunk 4]
```diff
@@ -189,88 +195,153 @@ def get_filepath_seasonal_lowertercile_raster(
...
+            "normal": da_no,
+            "above_normal": da_an.drop("quantile"),
+        }
+    )
+
+    da_season_terc.to_netcdf(chirps_seasonal_tercile_filepath)
+    return chirps_seasonal_tercile_filepath
+
+
+def _compute_bounds_terciles(da):
```
turnerm, line 323: A very small point, but to make the naming consistent I would call this `_compute_tercile_bounds`

turnerm (approved):
Amazing tests!

src/indicators/drought/chirps_rainfallobservations.py
Tinkaa: Haha I like it. Adjusted the formula a bit cause the `month_name` is 1-indexed. 
Anyway then realized, it would be even nicer to include the `seas_len` as parameter to it as well. I think the below makes sense but don't see how to do it in a list comprehension. Do you? 

``` python
end_month_season_mapping={}
for i in range(1,13):
    season_str=""
    for l in range(seas_len,0,-1):
        season_str+=month_name[(i - l) % 12+1][0]
    end_month_season_mapping[i]=season_str

Comment (turnerm) [reply src/indicators/drought/chirps_rainfallobservations.py:250]: I think something like this works: 
```python
{m: ''.join([month_name[(m - i) % 12 + 1][0] for i in range(3)]) for m in range(1, 13)}
```

Tinkaa: Love it!

Tinkaa:
Made one more change, namely to have the climatology years as parameter as these can change over time (and seem they will soon for the IRI forecast). 
Will merge now

PR #258, opened 2022-01-03
Feature/tcd cleanup

Tinkaa:
Clean-up the Chad risk data files by 
1) making sure the R and Rmd scripts work with the current data directory structure
2) removing unused images and scripts (mainly food insecurity scripts)
3) move used images to the GDrive

Thanks @joseepoirier for all the help. Since you created these scripts, I wanted to check with you if it is okay like this? 

Also, I would like to put the .Rmd and .html in a separate `docs` folder that is in the `tcd` folder but then I am getting import issues again. Do you know a solution for that?


joseepoirier:
Looping in @caldwellst as I believe he had looked into best practices for R projects. Seth, do you have a recommendation for the structure of Rproj folders? While at it, were you able to read more on library management in RProjs as well?

caldwellst:
Hey @joseepoirier! Yeah, I did a while ago and on this am quite familiar already with previous work. I posted this a while ago, but can't seem to remember where so will put again a few things that I think are best:

#### the [here](https://here.r-lib.org/) package

Whatever the directory structure, your `.Rproj` file should always lie at the top level of the project directory. Then, using the `here` package, you can use `here()` to construct OS non-specific, relevant paths from that top level. So, having your `.Rproj` file in `tcd` is probably the best choice here. Then you can arrange your scripts in whatever manner that you find relevant. If you have some that are being sourced by others, maybe because they have generic functions, you could place those in a specific folder separate from exploratory scripts and easily access them anywhere.

If you need access to the higher level `src` folder that we are using here, then that's another question, but that's why we are moving to project specific repos so would keep all of this in there anyway.

#### the [renv](https://rstudio.github.io/renv/articles/renv.html) package

This is the replacement for `packrat`. I will say again, I think for what we are doing, this might be overkill, because issues of backwards compatability in R are so much rarer than in Python so the issues are typically just installing a package that might be missing and we just rarely use R, but `renv` is simple and easy to use.

Just install the package, open up your R Project, and run `renv::init()`. Happy to help if you go this route and have any questions, but documentation is quite detailed and useful, see for instance the [init documentation](https://rstudio.github.io/renv/reference/init.html).

joseepoirier:
Thank you @caldwellst ! I did not know about `here` so I'll check it out. I agree that it sounds sufficient for our needs. 

Thanks for confirming `renv` is the successor to `packrat` and the best way forward. Library compatibility hasn't been an issue for us much (yet?). I used to conditionally install packaged in new scripts so anyone can run it regardless of their currently-installed libraries. Maybe it's a habit we should (re)commit to? 

If `renv` is an overkill for us, this will certainly be but in case you did not know about the tool: https://eliocamp.github.io/codigo-r/en/2021/08/docker-renv/

Tinkaa:
Ohh this is awesome, it works, thanks @caldwellst ! 

@joseepoirier does it content-wise look good to you?

joseepoirier:
@Tinkaa It does!

PR #259, opened 2022-01-04
analysis dec obs trig

Tinkaa:
Analysis on the precipitation and dry spells during December 2021 as requested. 

Could you check if it makes sense and if you got better ideas to visualize the storyline?


caldwellst:
Been getting continuously set back by tiny things in processing code, annoying! Will check this first thing in the morning and then we can send on to Josée for comment and discussion, so around 1-2PM your time will be done!

Tinkaa:
Nice, I like them!

Tinkaa:
Thanks for this Seth! I think we are done, so will merge

PR #260, opened 2022-01-05
Feature/tcd doc

Tinkaa:
Start of documentation in ` analyses/tcd/docs/tcd_trigger_development_summary.Rmd`

No need for real review yet, but more for your info. Especially tried to justify the list of drought years


caldwellst:
Hello hello! So, I've got all the figures up for analysis. However, think would be good to discuss the `.Rmd` and what we're looking for in it. Was adding to it but felt like it was getting a bit long and figured we could agree what we want in and out and then trim it for consumption before I plump it up too much!

Tinkaa:
@caldwellst this PR became a mess, but I think all of it has been reviewed in separate PRS. So is it okay for you to merge?

caldwellst:
Hey @Tinkaa, yeah, we've thrown so much together, but yeah, let's just merge it, I think is alright.

PR #261, opened 2022-01-06
compute stats for non cds data

Tinkaa:
Small bug for the mwi predictive trigger where stats weren't recomputed for data from ecmwf directly. 
Doesn't change previous results as we were recomputing them manually in another script before


turnerm (approved):
Looks good! Maybe just add a short comment in the code explaining why you need the new`if` statement

PR #264, opened 2022-01-10
Explore Mozambique available data

caldwellst:
Very simple analysis. Not used for anything public so if you want to just approve, feel free, and if you do more, don't spend too much time.


Tinkaa (approved):
Awesome! I couldn't refrain myself from at least quickly looking through the html slides, super interesting. 
Didn't look at the code but two checking questions: 
- Is it okay to publicly share the data? As the html will be publicly available now
- Could you put the image of hotspots on the Drive instead of on Github?

caldwellst:
Done!

PR #268, opened 2022-01-12
Feature/eth ndvi

Tinkaa:
Quick code to create a graph of the NDVI status (as indication of drought) for ETH for possible inclusion in the HNO

For now, it is only about the notebook, have to clean up `ndvi.py` a bit. 
Added also the `ipynb` as it takes about 15 minutes to run the whole notebook. 

Could you have a look at it from a conceptual level? Added few questions to the notebook, but mainly if we think the way the data is processed represents the drought status well


analyses/eth/notebooks/exploration/eth_ndvi.md
```diff
@@ -0,0 +1,205 @@
...
+                             legend=True,# if i==num_plots-1 else False,
+                cmap=ListedColormap(ndvi_colors)#,levels=ndvi_bins
+         )
+        ax.set_title(f"{data_col} of % NDVI of median for {pd.to_datetime(str(d)).strftime('%Y-%m-%d')}")
+        ax.axis("off")
+```
+
+#### Aggregated to admin3
+Below the values per admin3 are shown. We use the same bins as [those used by USGS/FewsNet](https://earlywarning.usgs.gov/fews/product/448). 
+We can see the same pattern as we saw with the raw data, which is a good sign. The dissimilarity between the pattern beginning of December and beginning of January is even more clear from these plots. It remains an open question which we should use..
```
caldwellst, line 186: So you are deciding to just use a single month's plots? I quickly checked the [FewsNet seasonal calendar](https://fews.net/file/113527), looking at the northern/southern pastoral regions, could it be that the north was greener in November/December based on their September/October rains and the November/December southern pastoral rains are what is driving the change in January? Might not make sense to expect patterns to remain the same given how large the country is.

Can you use both maps and also have a map of the differences? This could be January to December differences to see the trend more clearly between the two, rather than the reader having to manually compare. You could also do month-to-month differences of all 3 if you wanted, or December/January differences with November, really depends on the narrative I guess.

caldwellst (approved):
All looks good! Happy to chat about what it is you're trying to show and what might best visualize that, but think it's basicaly almost there unless you wanted to show explicitly the differences.

caldwellst:
I also just read your intro paragraph, which I avoided before diving right in ha, so apologies for directly reviewing the `.md`!

analyses/eth/notebooks/exploration/eth_ndvi.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+To check, we do the same computations for the adm2 to see if the general patterns are the same. As we can see they indeed generally match though as expected the values at admin2 are more smoothed
+
+```python
+gdf_stats_adm2=aggregate_admin(da,gdf_adm2,pcode2_col,bins=ndvi_bins)
+```
+
+```python
+plt_ndvi_dates(gdf_stats_adm2,"median_binned")
```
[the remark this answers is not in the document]
Tinkaa: Yes agree. Adm3 was requested, but will discuss with Josée to fine-tune the request

analyses/eth/notebooks/exploration/eth_ndvi.md
Tinkaa: Yes also agree! Initially the request was just the NDVI right now but I thought that might not reflect the drought situation well as as you note some parts of the country already have their rains in Sep/Oct/Nov. And this if there was extremely low NDVI during that time I would think there might be a high chance there is still significant drought impacts.. I guess that is what I am really struggling with, whether the NDVI at one point gives a good indication of drought.. Especially in a country with different precipitation patterns

analyses/eth/notebooks/exploration/eth_ndvi.md
```diff
@@ -0,0 +1,205 @@
...
+ndvi_bins=[0,60,70,80,90,95,105,110,120,130,140,200]
+```
+
+### NDVI rasters
+
+
+The NDVI raster files are downloaded manuallly from [here](https://edcintl.cr.usgs.gov/downloads/sciweb1/shared/fews/web/africa/east/dekadal/emodis/ndvi_c6/percentofmedian/downloads/dekadal/). Code is almost ready to do this automatically but due to some issues and time pressure, this was preferred
+
+
+From the plots we can conclude two main points:
```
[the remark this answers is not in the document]
Tinkaa: Agree largely. And think in this case (compared to e.g. dry spells) it is okayish to aggregate to admin area. Though I still think admin boundaries are a very artificial way of aggregation

analyses/eth/notebooks/exploration/eth_ndvi.md
caldwellst: I think there's so many other factors, don't say anything. I would express these concerns explicitly, link to FewsNet ([is this the latest](https://fews.net/east-africa/ethiopia/food-security-outlook/october-2021)?), and make it clear what you are and aren't saying. Hopefully in country they can interpret the maps in a meaningful/useful way, but yeah, don't think we can really say much more without a lot of additional analysis and exploration.

Tinkaa:
Alrighty, cleaned up and now also `ndvi.py` should be working for both regions. 
FYI, this is the graph now shared with the country office (did have to be at adm3) 
afbeelding

src/indicators/drought/ndvi.py
```diff
@@ -85,19 +95,33 @@ def download_ndvi(
...
+            if year == start_year and dekad < start_dekad:
+                continue
+            if year == end_year and dekad > end_dekad:
+                continue
+            if clobber or [year, dekad] not in dts:
+                print(year)
+                _download_ndvi_dekad(region=region, year=year, dekad=dekad)
+
 
-    for year, dekad in zip(range(start_year, end_year), range(1, 37)):
```
Tinkaa, line 90: So sorry @caldwellst I think I told you to make this a zip but was totally wrong, so changed it back

caldwellst: Ah woops! I should've tested before changing!

Tinkaa:
@caldwellst I added the code for computing the %of dekads below threshold. And made some changes to `ndvi.py` to account for the changed filenaming from `pct` to `pctm` (arghh). Could you have a last look?

caldwellst (approved):
All of it looks good to me, but just quickly gave a once over to the python file. Nice!

PR #269, opened 2022-01-20
Feature/tcd iri notebooks

Tinkaa:
Nothing special, mainly cleaning up the notebooks and reflecting in the text what we know now


caldwellst:
All looks good to me. The code runs and checked the changed text and it isn't wild!

PR #270, opened 2022-01-20
Feature/tcd chirps bootstrap

Tinkaa:
Notebook to apply bootstrapping to test the confidence in the relation between observed below average seasonal precipitation and historical drought years. 

The results are not looking promising from a "not by chance" perspective. 

Could you check if the methodology makes sense?

### DIFF
--- a/analyses/tcd/notebooks/forecasts/tcd_bootstrap_seas_bavg_precip_observed.md
+++ b/analyses/tcd/notebooks/forecasts/tcd_bootstrap_seas_bavg_precip_observed.md
@@ -14,6 +14,7 @@ import numpy as np
 import math
 from mlxtend.evaluate import confusion_matrix
 from mlxtend.plotting import plot_confusion_matrix
+import seaborn as sns
 
 from pathlib import Path
 import os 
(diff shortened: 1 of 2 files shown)


analyses/tcd/notebooks/forecasts/tcd_bootstrap_seas_bavg_precip_observed.md
```diff
@@ -0,0 +1,165 @@
...
+
+from statsmodels.stats.proportion import proportion_confint
+import matplotlib.pyplot as plt
+
+rng = np.random.default_rng(12345)
+```
+
+```python
+iso3="tcd"
+country_data_exploration_dir = Path(config.DATA_DIR)/config.PUBLIC_DIR/"exploration"/iso3
```
turnerm, line 25: `config` is undefined here

turnerm (approved):
This looks great Tinka. I think the results shouldn't be too surprising... precision/recall of <50% is probably not performing better than random, which is indeed what you have shown. Maybe we can discuss in Slack what this means for the pilot.

One thing though, are you confident that the methodology used for the observed precip is equivalent to what would be used for the forecast? Happy to discuss if you'd like.

Tinkaa:
Thanks! 

Just added a bit more comparison, also comparing to the other source of Biomasse. 
I will start a discussion on Monday haha, it is too close to the weekend now

>One thing though, are you confident that the methodology used for the observed precip is equivalent to what would be used for the forecast? 
I am pretty confident that it won't be equivalent. As the forecasts have the extra probability parameter which makes patterns quite different I think. However, in the end the forecasts are attempting to predict the quantity of below average precipitation. So if below average precipitation doesn't show any relation with drought impact, then the rationale of using these seasonal forecasts is a lot weaker..

PR #271, opened 2022-01-25
analysis rainy onset in tcd

Tinkaa:
Exploration of the onset dates in TCD. 

Not critical for anything right now, but couldn't get rid of my curiosity to see if it shows different patterns than seasonal below average precipitation (which it does!)


caldwellst (approved):
All looks good and runs for me, except that weird issue with the datetime. I don't know what's happening here, happened to me before but then has seemingly gone away. Maybe we can create an issue to sort out?

As well, I've been thinking that we could have a set of methdos to do on various raster data. These could implemented as extensions to data arrays, like `rio`, but just for rainfall data. Then we could implement a lot of these various ARC2 metrics and they would be immediately available for ARC2, but also CHIRPS and any other precipitation data, such as calculating rainy season based on multiple definitions, calculating dry season onset, etc. Could be interesting!

PR #272, opened 2022-01-25
Uncertainty report

joseepoirier:
@turnerm This is the doc we would submit with a formal proposal and in the final AAF. It's mock data as we discussed and it would read in the data from csv files generated through analytical scripts.

I didn't push any images. You should be able to generate them as you knit the document. 

@caldwellst The CSS doesn't all work properly. Can you take a look with your fresh eyes to figure out why font color of headers aren't working, for instance?


analyses/country_template/docs/style.css
```diff
@@ -1,11 +1,34 @@
...
+h1 {
     font-family: Arvo;
     font-weight: bold;
     color: #1bb580;
+    font-size: 18px;
+}
+
+h2 {
+    font-family: Arvo;
+    font-weight:
```
caldwellst, line 14: Empty CSS styling is preventing the color from being read properly below.

caldwellst (commented):
Just checked quickly CSS ha, just a simple change.

analyses/country_template/docs/style.css
joseepoirier: Yesss! Thanks for the fresh eyes, Seth!

analyses/country_template/docs/trigger_report_template.html
```diff
@@ -0,0 +1,622 @@
+<!DOCTYPE html>
```
turnerm: If this is generated, then maybe it doesn't need to be checked in?

turnerm: If this is generated, then maybe it doesn't need to be checked in?

turnerm (approved):
Looks really great, thanks for creating this Josée!

analyses/country_template/docs/trigger_report_template.html
joseepoirier: I figured this way it is easy to share with outside, non-technical folks. But I guess this is the template so we wouldn't share an "empty" template. I'll remove it.

analyses/country_template/docs/dummy_perf_metrics_table_for_template.csv
```diff
@@ -0,0 +1,61 @@
+﻿trigger,metric,point,value,dummy
+a,var,low_end,85,1
```
turnerm, line 2: Do we expect the input to be rounded to an int, or can the user input floats?

turnerm (commented):
I'd like to play around with it a bit more, but I'm getting an error when I try to knit now

turnerm:
> The values could be floats. I wouldn't go beyond 1 decimal however, would you? I also haven't tested the visuals with floats so the labels on the CI plots might get a bit crowded. A first question we should answer is whether we think it is necessary and justified to report at the decimal level, esp since they are values within a confidence interval (not means being compared in a t-test, for instance - which requires high precision)?

@joseepoirier  Agree that we definitely shouldn't be reporting floats. I just wanted to check whether the rounding would occur here, or if the user should pre-round the values for the input csv.

joseepoirier:
I originally had it in the Rmd but removed it, thinking it would be best done as part of a set of validation checks when generating the values. Happy to reconsider though: it's only a few lines of code to add if needed.

It also shouldn't be the reason you can't knit the document, in case that was your line of thought.

analyses/country_template/docs/trigger_report_template.Rmd
```diff
@@ -0,0 +1,333 @@
...
+funding_a <- paste0('$',funding_a,'m')
+funding_b <- paste0('$',funding_b,'m')
+funding_c <- paste0('$',funding_c,'m')
+
+```
+
+```{r compute CI measurements}
+
+# compute width and position of CIs for trigger a
+df_a_var <- computeCIwidths(metric_name = 'var', trigger_id = 'a') # valid activation rate
```
turnerm, line 132: I think this section can be simplified with `map2_dfr`, something like: 

```r
df_all = purrr::map2_dfr(
  rep(c("var", "det", "far", "mis", "atv"), times=3),
  rep(c("a", "b", "c"), each=5),
  ~ computeCIwidths(metric_name = .x, trigger_id = .y)
)
```

I think you would need to add the trigger ID to the data frame produced with `computeCIwidths`

analyses/country_template/docs/trigger_report_template.Rmd
```diff
@@ -0,0 +1,333 @@
...
+df_fmwk_min <- computeCIwidths(metric_name = 'min', trigger_id = 'framework') # prob minimum framework activation 
+df_fmwk_ful <- computeCIwidths(metric_name = 'ful', trigger_id = 'framework') # prob full framework activation 
+
+```
+
+```{r data plotting}
+
+source("computeCIplots.R")
+
+plotCI(trigger_id = 'a', metric = 'var', ci_widths_df = 'df_a_var')
```
turnerm, line 162: I think you could also use `map` here to simplify

turnerm (commented):
I solved the error that I was getting: it was due to there not being a `plots` directory in `docs`. Would it be possible to add the creation of this directory (if it doesn't exist) to the code?

joseepoirier:
@turnerm Ha yes. It's indicated on ln35 that the folder needs to exist. I didn't set it up to do it automatically because the repos are undergoing so many refactorings and restructuring, I didn't know if this were the path or structure we'd want. Figured it was a non-issue until we start using it and we can make a decision then.

analyses/country_template/docs/trigger_report_template.Rmd
joseepoirier: Yes indeed. I didn't refactor the code and left it as explicit and easy to read as possible for non-R users. I also wanted the documentation to  be clear whereas the map2_dfr is more opaque.

turnerm: Do you expect a lot of non-R users to be using this? I assume it will mostly be us (using the source code at least), so I would advocate for making our own lives more straightforward, especially if we're going to be making several copies of this document. The quick flexibility of `map` would be nice in that case.

turnerm:
> I didn't set it up to do it automatically because the repos are undergoing so many refactorings and restructuring, I didn't know if this were the path or structure we'd want. Figured it was a non-issue until we start using it and we can make a decision then.

Ah okay makes sense. I would argue for setting it up so at least the current version of the rmd file will run smoothly for whoever is using it; it can always be changed if we want to modify the file structure.

analyses/country_template/docs/trigger_report_template.Rmd
joseepoirier: Same as above

joseepoirier: Removing this change because the plotCI function doesn't handle lists well. Will look into ti when we revisit this Rmd to reflect feedback from colleagues.

joseepoirier:
@turnerm Per our conversation I'll merge this for now and do some refactoring later on, perhaps to reflect feedback as it comes in.

PR #275, opened 2022-02-08
Feature/ssd floodscan

Tinkaa:
First exploration of the floodscan data for selection of counties in Jonglei and Unity (SSD). 

Not super clean but wanted to do it quick and iteratively. So mainly conceptually, what do you think of the approach? Do you think the list of flood events per county make sense or there is a better approach here? Happy to call as well!


turnerm (commented):
Looks great so far, but pausing the review to sort out an error first

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,324 @@
...
+
+```python
+fs_clip_rainy.hvplot.hist('SFED_AREA',alpha=0.5)
+```
+
+```python
+fs_clip_rainy.where(fs_clip_rainy>0.4).hvplot.hist('SFED_AREA',alpha=0.5)
+```
+
+When only focusing on the states of Jonglei and Unity, we can see that on average they show slightly higher values, though the majority of the values is still below 0.2
```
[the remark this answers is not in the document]
Tinkaa: >the pixels should really be used in aggregate

What do you mean? That we should have the original higher resolution data?

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,324 @@
...
+#         lon_coord="lon",
+#         lat_coord="lat",
+#         stats_list=["median","mean","max","count"], #std, count
+#         #computes value where 20% of the area is above that value
+#         percentile_list=[80],
+#         #Decided to only use centres, but can change that
+#         all_touched=False,
+#     )
+# df_floodscan=df_floodscan.merge(gdf_adm2[["ADM2_EN","ADM2_PCODE"]],on="ADM2_PCODE",how="left")
+# df_floodscan.to_csv(country_data_exploration_dir/'floodscan'/f'{iso3}_floodscan_adm2_stats.nc',index=False)
```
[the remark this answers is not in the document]
Tinkaa: Arghh something icky with the crs is going on when saving the raster data. Not perfect fix, but it should work now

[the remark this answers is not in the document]
turnerm: Yep it's running now thanks!

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,327 @@
...
+
+When only focusing on the states of Jonglei and Unity, we can see that on average they show slightly higher values, though the majority of the values is still below 0.2
+
+```python
+(fs_clip_rainy.rio.write_crs("EPSG:4326").rio.clip(gdf_adm2.loc[gdf_adm2.ADM1_EN.isin(['Unity','Jonglei']),'geometry'])
+.hvplot.hist('SFED_AREA',alpha=0.5))
+```
+
+### Compute stats per county
+Now that we have analyzed the raster data, we can aggregate this to get the statistics per county.
```
turnerm, line 124: I have to look more closely to figure out what you're calculating, but I need to go soon so before I do I just wanted to mention something. 

As I said before, I'm wondering if it's better to focus on total flooded area rather than the flood fraction of individual pixels. since I imagine there's probably quite some uncertainty at the pixel level. And even starting with looking at country-level rather than breaking it down into counties, which could bias the results. 

So what I mean is just for each day just summing the pixel fractions, and then dividing by the total number of pixels to get the amount of flooded area. From that you could identify when the major floods occur, and then see where they are happening geographically. Hope that makes sense.

Tinkaa: Hmm I like the thinking. However, rethinking it I think that is exactly what I am doing when taking the mean of the pixel fractions. 
mean(flood fraction of pixels within county) = sum(flood fraction of pixels within county)/#pixels within county. 
So actually instead of refering to the mean of the pixel fractions, it could refer to the total fraction of the county (or country) that is flooded. Do you agree or am I talking late afternoon bullshit here?

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,324 @@
...
+
+From the plot below we can see that the fraction of the area that is flooded (SFED_AREA) is generally very low, barely reacing above 0.2. I am not sure why this is the case. 
+Note that from the first graph you cannot see the higher values well. It does occur that there are values higher than 0.2 but much less often, see the second graph. 
+
+```python
+fs_clip_rainy=fs_clip.sel(time=fs_clip.time.dt.month.isin(range(7,11)))
+```
+
+```python
+fs_clip_rainy.hvplot.hist('SFED_AREA',alpha=0.5)
```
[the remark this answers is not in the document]
Tinkaa: I like the logging! But seems there is an [unsolved issue of hvplot](https://github.com/holoviz/holoviews/issues/2591) not doing that correctly. Filtered values below 0.01 now.

caldwellst:
No review yet, but Monica commenting on the code probably sufficient anyway! Overall, analysis looks fantastic, thanks. Just some thoughts from my side:

- Would be interested to look at the timeline of flooding geographically to see if there's anything relevant there. I quickly threw together a graph before I have to run now, looking at the counties west and east of the Nile (by column) and organized north/south. I quickly smoothed, so might not be optimal, but seems to show that at least in 2020 there's a long build up of flooding, particularly in the central wetlands, where peaks might be reached earlier than in counties further north/south.
image

- On top of that, can't be seen there, but wondering if over the years we see certain counties reaching peaks earlier, or trends starting earlier, depending on location? For instnace, if we can see it moving south/north along the Nile, would be great to incorporate in the trigger design.
- Non-county-wise, for looking at the overall lead time we get, maybe start at the national level compared to those downstream water sources + rainfall and lagged flood data, then we can get into details on the counties. I'm sure that was the plan, just came to my mind.

Anyway, would love to chat about this, really cool work, but gotta run, so let's be in touch tomorrow. Thanks a lot Tinka!

analyses/ssd/flood_trigger/ssd_floodscan.md
turnerm: > am I talking late afternoon bullshit here?

:joy: not at all, in fact you're totally right! 

Two more things that I was thinking about: 

1. Do you know how accurate the pixel fractions are? I haven't read any of the docs, but I'm wondering if pixel fraction is more either (a) a confidence on whether a given pixel is flooded or not, or (b) can sort of be ignored and we should just take any pixel > 0 as flooded
2. Probably out of scope but I'm wondering if we need to take into account clustering (spatial, but we could also do temporal to find events)

Tinkaa: 1. As far as I understand is that the original resolution is 3 arcsec. At this resolution, the data is binary i.e. 1 is flooded 0 is non-flooded. Now the data we are using is 300 arcsec resolution. So I think (but this is not documented) that the fraction is just the fraction of the 1000 original pixels in our 1 pixel that are flooded. Does that make sense to you? Could double-check with the floodscan team
2. You mean if pixels close to each other show similair patterns?

caldwellst: Yeah, was thinking the same thing about the clustering. Wondering if we have access to the `NDAYS` product described in the user documentation or if we can just produce this is ourselves?

image

And interesting Tinka, definitely makes sense to me regarding the resolution/product.

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,372 @@
...
+fs_clip.NDT_SFED_AREA.attrs.pop('grid_mapping')
+fs_clip.LWMASK_AREA.attrs.pop('grid_mapping')
+fs_clip=fs_clip.rio.write_crs("EPSG:4326",inplace=True)
+```
+
+We plot the data. We can see that the resolution is high (300 arcseconds = 0.083 degrees). We therefore decide that when analyzing the data at the admin2 level, it is fine to only look at the cells with their centre within the admin2. 
+
+```python
+g=fs_clip.sel(time='2020-08-03T00:00:00.000000000').SFED_AREA.plot()
+gdf_adm2.boundary.plot(ax=g.axes);
```
turnerm, line 89: Btw here I was playing around a bit with `hvplot` and `groupby='time'` to get a scroll bar. I couldn't quite get it to work though.

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,372 @@
...
+        ax.set_title(f'Flooding in {adm}, 1998-2020, aggregation {var_col}')
+        ax.legend()
+```
+
+```python tags=[]
+plot_floodscan_timeseries(df_floodscan,"mean_ADM2_PCODE",['Fangak'])
+```
+
+```python
+plot_floodscan_timeseries(df_floodscan,"mean_ADM2_PCODE",adm2_list)
```
turnerm, line 178: The data from Leer looks very strange to me, how the baseline suddenly changes halfway through:
download

Also Twic East and Bor South are a bit weird, it's more difficult to see the seasonality:
download
download

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,372 @@
...
+```python
+df_year=df_floods_all_median[['ADM2_EN','year']].drop_duplicates()
+```
+
+```python
+df_year.groupby("ADM2_EN").count()
+```
+
+```python
+df_year.groupby("year").count().sort_values('ADM2_EN',ascending=False)
```
turnerm, line 287: This seems like a large number of events. Are we aiming for a particular return period? Nationally or per county? Since 3 standard devations is 68% (assuming normality...) then I think that would correspond to 1/(1-0.68)= 1 in 3 year RP (per county). 

I guess I'm starting to worry that chopping things up by county is going to create some weird artefacts in the analysis. There are definitely counties that only have a few pixels.

analyses/ssd/flood_trigger/ssd_floodscan.md
```diff
@@ -0,0 +1,372 @@
...
+    x="year:N",
+    y="ADM2_EN:N",
+     color=alt.Color('flood_event:N',scale=alt.Scale(range=["#D3D3D3",'red']))
+).properties(
+    title="Flood detected per county"
+)
+```
+
+### Flooded fraction in the country
+As we can see there is a lot of fluctuation between the counties. Instead we can also look at the total flooded fraction in the country, as shown below.
```
turnerm, line 307: Ah yes, I would definitely go this route, and maybe calculate RP instead of using SDs

turnerm (commented):
Really good Tinka, I guess my one suggestion is to use RPs instead of SDs (or maybe a higher SD threshold which is closer to the desired RP). 

As for county- vs full country-level, I guess it depends on how the targeting will work, maybe @caldwellst has more to say about it. Another option is to use a somewhat lower RP for full country level, but then every time there is a flood, check whether it impacts the counties we are interested in. Hope that makes sense.

caldwellst:
> Really good Tinka, I guess my one suggestion is to use RPs instead of SDs (or maybe a higher SD threshold which is closer to the desired RP).
> 
> As for county- vs full country-level, I guess it depends on how the targeting will work, maybe @caldwellst has more to say about it. Another option is to use a somewhat lower RP for full country level, but then every time there is a flood, check whether it impacts the counties we are interested in. Hope that makes sense.

Targeting is not clear yet, but given the scale of the response in South Sudan, I think that it will be quite focused on specific areas that are operationally feasible, have high levels of vulnerability or poor current humanitarian situation, and for which we have some signalling ability.

For the start of the analysis, I think might be best just to look at a wider level than the counties, possibly a collection of those seeing this flooding? Or could be nationwide. This is just a starting off point to see if there appears to be any useful signal using rainfall and hydrological data, because if not, we might resort to focusing the analysis more on area selection to ensure that even with a bad signal, we're at least reaching the right populations.

Tinkaa:
Loving these conversations! 

Really cool graph of 2020 Seth. I tried to see if there is a general pattern of counties reaching flood peak earlier but there doesn't seem to be that much signal (which results in a very messy graph haha). Though I am very doubtful if flood peak is the right measure given the fluctuations in the data (the `N_DAYS` is probably better in that perspective).

I agree that counties is artificial. Was also thinking of chopping it up into rectangles instead of counties. But I think the struggle is that we should somehow identify first where flooding even takes place.. 
@turnerm that is why I used SD instead of RP. With RP, you would kind of assume every county sees flooding, which I wasn't sure of. But then again I think the SD isn't working very well either. 

But maybe if we just resort to the country level we can skip the question of in which areas we should "measure" the flooding a bit

afbeelding

analyses/ssd/flood_trigger/ssd_floodscan.md
Tinkaa: See conversation below but yes, 3SD was meant as 1 in 3 year return period, but didn't want to force a 1/3 RP since we don't know if flooding occurs in all counties. 
However, since we compute SD on all data within the rainy season, I don't think it is the best method either cause it means it will indicate floods more often in counties that have a lot of spikeness.. Do you have thoughts on what to best measure? 

>I guess I'm starting to worry that chopping things up by county is going to create some weird artefacts in the analysis.

Agree

analyses/ssd/flood_trigger/ssd_floodscan.md
Tinkaa: Nice catches! Maybe we can ask the floodscan people about them.

analyses/ssd/flood_trigger/ssd_floodscan.md
Tinkaa: We don't have the NDAYS :( And maybe a bit harder to produce given we don't have the binary data, but we could do smth like NDAYS>pixel fraction 

For the clustering, makes sense!

turnerm: > As far as I understand is that the original resolution is 3 arcsec. At this resolution, the data is binary i.e. 1 is flooded 0 is non-flooded. Now the data we are using is 300 arcsec resolution. So I think (but this is not documented) that the fraction is just the fraction of the 1000 original pixels in our 1 pixel that are flooded. Does that make sense to you? Could double-check with the floodscan team

Good to know, thanks for explaining! 

> You mean if pixels close to each other show similair patterns?

Yeah, I would probably be inclined to define a "flood" as when many adjacent or nearly-adjacent pixels are > 1. If there are just random pixels > 1 with no connection to others then I would guess they are more likely to be FP.

caldwellst:
Thanks Tinka for the GLOFAS update. I think is interesting for sure, but the main problem is essentially we are looking at river discharge data downstream from the Sudd system, where the vast majority of flood dynamics are seen. In particular, in 2021, given the already high levels of flooding, and that they hadn't yet receded back into the Nile basin but remained in surrounding areas, new flows into the area during the rainy season overwhelmed surrounding areas creating the massive flood extents, and these still have yet to significantly drain.

This could be why the discharge downstream is so low, because the traditional mechanisms for the Nile through-flow have been disrupted, so all excess water is just spilling out rather than flowing through and being registered as a spike in GLOFAS/the Malakal station. This is broadly what some of the flood experts have been saying around here, and one mentioned that this overflow essentially created spill over into the W/NBeG river basins that additionally overloaded those rivers and caused flooding across areas like northern Unity not directly related to the Nile/Sudd basin.

Tinkaa:
Thanks Seth, super interesting stuff! 
Tots agree that Malakal wouldn't be the right place to look as it is so downstream and your explanation makes a lot of sense why discharge wasn't high at the years we saw large floods. 
It doesn't fully explain to me the phenomenon the other way around, i.e. we saw large river discharge but no floods. But well, as Monica mentioned before, the dynamics between when river discharge causes higher water levels causes flooding are very complex. Let alone that we have no idea how reliable GloFAS is at Malakal. Plus too short leadtime and yeah well no-way it is really useful.

analyses/ssd/flood_trigger/ssd_floodscan.md
Tinkaa: Long time ago, but yes agree. Looking at RP per year makes more sense

analyses/ssd/flood_trigger/ssd_floodscan.md
Tinkaa: Reply from Floodscan on Leer: I haven’t found any obvious causes for the pattern in Leer. There are a few possibilities. An increase in vegetation cover over time might cause less flooding to be detected. However, I don’t see an obvious greenness trend over that time period. (I used NASA’s WorldView tool to browse through vegetation index in July each year when flooding is usual at its peak: https://worldview.earthdata.nasa.gov/). Changes in the sensors available to us might have an effect but those changes (in Oct. 2011, Oct. 2012, and April 2014) don’t seem to correlate to the trends in Leer either. Otherwise, I can only speculate. I suppose it’s possible that the floodplain evolved and silting or other factors reduced flooding in some areas for a time. More likely is that some of the area is on the margins of the seasonal flood and so has more extreme year-to-year variation in flood extent.

analyses/ssd/flood_trigger/ssd_floodscan.md
Tinkaa: Yes! Added a note about it.

PR #276, opened 2022-02-21
Feature/ssd glofas

Tinkaa:
Comparison between reanalysis glofas data in Malakal and floodscan data for that county. 
Goal was to see if there was a similair pattern between the two (which there isn't really but might be partly due to strange floodscan data). 

I don't fully understand yet how river discharge and flooding in theory are supposed to relate, so after you have had a look, maybe we could have a call to discuss?

### DIFF
--- a/analyses/ssd/flood_trigger/ssd_glofas_vs_floodscan.md
+++ b/analyses/ssd/flood_trigger/ssd_glofas_vs_floodscan.md
@@ -15,6 +15,8 @@ However, the peak years of the GloFas data don't fully correspond with the peak 
 This notebook is inspired by the Malawi glofas vs floodscan analysis
 <!-- #endregion -->
 
+![afbeelding.png](attachment:673c27fd-83fb-4679-aa6d-a7d53b1703ee.png)
+
 ```python
 import matplotlib.pyplot as plt
 import matplotlib as mpl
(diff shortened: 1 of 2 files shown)


turnerm (approved):
> I don't fully understand yet how river discharge and flooding in theory are supposed to relate, so after you have had a look, maybe we could have a call to discuss?

Yes great question. In principle, river discharge should correspond to water level according to a [rating curve](https://en.wikipedia.org/wiki/Rating_curve). And when water level exceeds the height of a river bank, flooding occurs. However, after that point the relationship breaks down because it depends on how the water spreads around etc. Also, the connection between water level and flood extent is rather complicated for the same reasons. 

At any rate, your analysis looks very well done. The only thing you could maybe add more is just computing the pearson's correlation (it should be quite bad though)

But yeah, going back to the connection between the flooding extent and river discharge. While the relationship is complicated for the reasons I mentioned above, I would naively expect to see  perhaps a lagged excess in flood extent after a river discharge peak. I find it a bit odd that the annual floodscan peak appears to come before the GloFAS one.

The difficulty in this case is that we aren't sure that either dataset is trustworthy.

Tinkaa:
Belated thanks! Added the correlation (which is horrible). If we do use Floodscan the spikey pattern in Malakal is def smth I would want to ask the team about.

PR #280, opened 2022-03-04
Feature/ssd fsnms needs

caldwellst:
I received data from the FSNMS+ (essentially MSNA for the HNO) from the IOM team, because they ran the urban site assessments in Bentiu IDP Camp and surrounding informal displacement sites. The idea was to explore if we could identify any clear differences in need between the two areas to help contextualize the suggested response plans coming from operational agencies.

The issue is, the data is garbage. I calculated the food consumption score (FCS) and then just quickly did some checks on it and the sad reality is the data at least for FCS is absolutely skewed. While the informal site populations have a near-normal distributed FCS, the skew in the camp populations is extremely large. As well, the difference in severe scores (< 21) would show the camp being magnitudes worse than the informal sites, which is just so unlikely it couldn't be true (heavily served camp, permanent populations vs. newly flood displaced populations living in informal sites). So, we are giving up on using the data because it's clearly not been managed or cleaned.

Not sure there is much we want to do with the analysis, so wouldn't necessarily bother reading through, but putting here for posterity.

### DIFF
--- a/analyses/ssd/flood_trigger/ssd_rubkona_needs.R
+++ b/analyses/ssd/flood_trigger/ssd_rubkona_needs.R
@@ -1,0 +1,131 @@
+library(tidyverse)
+
+# Explore the latest FSNMS data that covers Bentiu town
+# and surrounding Bentiu IDP sites. This can be used
+# to inform the potential selection of activities across sites
+
+file_dir <- file.path(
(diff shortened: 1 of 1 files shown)


Tinkaa (approved):
Sad, but smart checks you did

PR #282, opened 2022-03-07
Feature/ner food insecurity

joseepoirier:


joseepoirier:
@Tinkaa I thought this branch had been merged but apparently not. Should we?

joseepoirier:
I actually didn't go through with the analysis (it was out of curiosity), but the docs could be used to pull FS data in the future. If not worth it we can also just delete everything.

Tinkaa:
Awesome you used the HDX R library!

We are not planning on doing any additional analysis on NER right?

Cause in that case I would tend towards deleting it 😢 Reasoning being that merging this will create the idea that we did analyses for NER while we didn't really. And in principle how to pull FS data will be saved by the legacy of this PR

(another option would be to use the same methodology for the bfa/tcd FS data but I get if you don't feel like investing in that) 

Would you agree to not merge this branch?

joseepoirier:
Yeah I thought we might go that route since the work wasn't fully executed or used. I'm okay with deleting the branch, I mainly wanted to clean it since it's been open for a while.

Tinkaa:
Sounds good!

PR #292, opened 2022-04-27
Ssd analysis

caldwellst:
Hi Tinka, you can find in the new file the code to produce the graph in the blog. All that is needed now is to update the graph once new data comes in from Floodscan. Thanks for helping with this while I'm away! Putting as draft since we still need the data and then to update the processing code when it arrives, which can go in this PR.


Tinkaa (approved):
I added the new floodscan data and imo everything looks good.

PR #293, opened 2022-05-06
Update README.md

joseepoirier:
The map wouldn't display. Updated the URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ While the triggers are adjusted to the local context, most of the data processin
 
 The AA frameworks that are implemented and/or currently under development are displayed in the map below
 
-![image](![image](https://user-images.githubusercontent.com/4512594/167153028-8c828a48-13a7-4625-84ad-b68750f0e942.png))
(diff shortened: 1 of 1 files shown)


joseepoirier:
@turnerm Ok to merge this?

turnerm:
Oh my gosh, so sorry I missed this somehow! In [the preview](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/7f36842dbb7aa7a4199c8956615976969daf1a3d/README.md) the image is still broken for me. Are you sure you need both `[!image]`s?

turnerm:
No changes from my end! But I'm so sorry, I still don't see the image... I can try to fix it next week

joseepoirier:
Hmm. Weird. I wonder if it's because GH is referring to a "GH user" link instead of the original URL I supplied? That would be silly but it's strange that a public image doesn't show up for other uses in a public repo.

This isn't urgent. Was just trying to clean up open PRs and branches. 

Another thought: I wonder if Github now allows iframes in README's? If so we could embed the AA webpage map here so we don't have to sync up and update the two...

turnerm (approved):
@joseepoirier don't hate me but I just checked again because I was going to try to fix it, and it works for me now...

PR #295, opened 2022-05-17
New district for Nepal

turnerm:
A check to see if flooding "events" are correlated with a proposed targeting district in a new river basin. No rush with this review.

### DIFF
--- a/src/npl/config.yml
+++ b/src/npl/config.yml
@@ -66,3 +66,8 @@ glofas:
     Simle_v3:
       lat: 27.05
       lon: 87.15
+    # DHM point added later
+    # https://hydrology.gov.np/#/basin/4635?_k=ep79ty
+    Kandra:
+      lat: 28.7
+      lon: 80.9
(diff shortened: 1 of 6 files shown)


src/npl/config.yml
```diff
@@ -66,3 +66,8 @@ glofas:
     Simle_v3:
       lat: 27.05
       lon: 87.15
+    # DHM point added later
```
Tinkaa, line 69: would add here a comment why the coordinates are not 0.05 rounded and when using the use_incorrect_coords add a comment that the coords are not rounded and thus using this. As you wouldn't expect this to be the standard

Tinkaa (approved):
Sorry for the delay.. Looks good and everything runs! Just add explanation of why using incorrect_area_coords

Tinkaa:
Oh and, there is no explanation to the analysis you added. Don't know if it was really like comparison, but if so I would add few sentences on the differences and why you chose one over the other if that is the case. (e.g. it seem there is a consistent bias but also some differences in patterns at peak moments)

PR #296, opened 2022-05-27
Chirps analysis

PaulineNimo:


analyses/npl/14_chirps_analysis.Rmd
```diff
@@ -18,106 +18,489 @@ library(tidyquant)
...
+    # close
     nc_close(ncObj)
-    return(DataCube)
+    # data object
+    return(bind_rows(KarDataCube, KoshDataCube))
 }
 
+
 # chirps data frame
 chirpsData <- list.files(filePath, pattern = "*.nc") %>%
```
turnerm, line 96: Oh wow, this does take a _really_ long time. You don't have to change it in case you never need to run it again, but in case you do, some ideas would be:
- Use `load()` in the aa-toolbox to create a single file 
- Similarly, use [this](https://rdrr.io/github/RS-eco/processNC/man/mergeNC.html) R function

analyses/npl/14_chirps_analysis.Rmd
```diff
@@ -18,106 +18,489 @@ library(tidyquant)
...
+nc_close()
+
+NPLWaterLevel <- read_csv(paste0(Sys.getenv("AA_DATA_DIR"), "/private/exploration/npl/dhm/processed/waterl_level_procssed.csv"))
+
 # function for reading in files 
 read_files <- function(filePath, fileName){
     ncObj <- nc_open(paste0(filePath, fileName))
-    DataCube <- ncvar_get(ncObj, "precipitation") %>%
+    # Karnali
+    KarnaliBox <- nplRBSF %>%
```
turnerm, line 48: Because the code for KarnaliBox and KoshiBox is the same, I would generalize it to a function where you can just pass a shapefile object, and then call it for each one.

analyses/npl/14_chirps_analysis.Rmd
```diff
@@ -18,106 +18,489 @@ library(tidyquant)
...
+    
+    KarBoxXMin <- which.min(abs(KarnaliBox[1] - round(ncvar_get(ncObj, "X"), 3)))
+    KarBoxXMax <- which.min(abs(KarnaliBox[2] - round(ncvar_get(ncObj, "X"), 3)))
+    KarBoxXLen <- KarBoxXMax - KarBoxXMin + 1
+    KarBoxYMin <- which.min(abs(KarnaliBox[3] - round(ncvar_get(ncObj, "Y"), 3)))
+    KarBoxYMax <- which.min(abs(KarnaliBox[4] - round(ncvar_get(ncObj, "Y"), 3)))
+    KarBoxYLen <- KarBoxYMin - KarBoxYMax + 1
+    KarDataCube <- ncvar_get(ncObj, "precipitation", 
+                             start = c(KarBoxXMin, KarBoxYMax, 1), 
+                             count = c(KarBoxXLen, KarBoxYLen, 1)) %>%
```
turnerm, line 60: Do I understand correctly that you're using the square bounding box? In principle we want to use the actual shape if possible. But if that's a bit tricky (I have no idea how to do it in R, aha) then the box should be fine for this work.

turnerm: @PaulineNimo just wondering, can you confirm what it is you're doing in this analysis -- is it using a box shape?

turnerm (approved):
Looks great, thanks for making the changes, sorry I was so slow to review!

PR #299, opened 2022-07-07
Discharge Data vs GLOFAS

PaulineNimo:


analyses/npl/16_dischargevsGLOFAS.Rmd
```diff
@@ -0,0 +1,167 @@
+---
+title: "DischargevsGLOFAS"
+output: md_document
+---
+
+```{r intro, echo=FALSE, warning = FALSE, message=FALSE, results='hide'}
+filePath <- paste0(Sys.getenv("AA_DATA_DIR"), "/public/processed/npl/chirps/")
```
turnerm, line 7: I don't think you need this line

analyses/npl/16_dischargevsGLOFAS.Rmd
```diff
@@ -0,0 +1,167 @@
...
+        full_join(DF2, by = "Basin") 
+    FloodCheck[,"TimeDiff"] <- difftime(FloodCheck[[DateCol1]], FloodCheck[[DateCol2]], units = "days")
+    FloodCheck[,"Check"] <- ifelse(FloodCheck$TimeDiff >= -15 & FloodCheck$TimeDiff <= 15, T, F)
+    
+    return(FloodCheck)
+}
+
+returnValues <- NPLFlow %>%
+    group_by(Basin) %>%
+    summarise(returnyr2 = return.level(fevd(x = Flow, method = "MLE", type="GP",
```
turnerm, line 75: I would use GEV instead of GP. Let me know if it changes the results significantly.

analyses/npl/16_dischargevsGLOFAS.Rmd
```diff
@@ -0,0 +1,167 @@
...
+    arrange(Dates, Basin) %>%
+    left_join(returnValues, by = "Basin") %>%
+    mutate(thresh2Yr = Flow >= returnyr2) %>%
+    mutate(thresh5Yr = Flow >= returnyr5) %>%
+    mutate(thresh10Yr = Flow >= returnyr10) %>%
+    filter(thresh2Yr)
+```
+
+```{r}
+returnValues2 <- NPLGLOFAS %>%
```
turnerm, line 91: maybe call this one returnValuesGlofas instead of 2 (and the other one DHM)

analyses/npl/16_dischargevsGLOFAS.Rmd
```diff
@@ -0,0 +1,167 @@
...
+    group_by(Station) %>%
+    summarise(returnyr2 = return.level(fevd(x = GLOFAS, method = "MLE", type="GP", 
+                                            threshold = 6000), return.period = 2), 
+              returnyr5 = return.level(fevd(x = GLOFAS, method = "MLE", type="GP", 
+                                            threshold = 6000), return.period = 5), 
+              returnyr10 = return.level(fevd(x = GLOFAS, method = "MLE", type="GP", 
+                                             threshold = 6000), return.period = 10)) %>%
+    mutate(Basin = ifelse(Station == "Chisapani", "Karnali River Basin", "Saptakoshi Watershed"))
+
+AggGLOFAS <- NPLGLOFAS %>%
```
turnerm, line 102: Do you remove dates that are outside of the DHM date range?

turnerm (approved):
Looks great, just a couple of questions / comments!

analyses/npl/16_dischargevsGLOFAS.Rmd
PaulineNimo: No, I didn't

turnerm: Ah, that is very important to do, otherwise the number of GloFAS events will be much higher. The date ranges need to be fully overlapping for the TP / FP / FN calculation.

analyses/npl/16_dischargevsGLOFAS.Rmd
turnerm: I just read the explanation, and now I understand the confusion: for the RP we only want to use the maximum each year, and not consider multiple events per year. So in that case GEV should work, I believe.

turnerm: Great thanks, let me know if this changes the results / conclusions!

turnerm: Should be good to merge then!

PR #301, opened 2022-07-18
update cartopy

Tinkaa:
I had some import issue when trying to run the code. It might have been my installation, but with updated cartopy I didn't have issues anymore so thought it wouldn't harm to update it in the requirements.


Tinkaa:
Ah okay! I can also check after holiday if 0.20.2 works for me. For now just use what works for you :)

Tinkaa:
Changing this back to 0.20.2

PR #303, opened 2022-07-28
Get GlofAS stations

turnerm:
Quick script to get GloFAS station for toolbox config. No rush for review.

### DIFF
--- a/src/indicators/flooding/glofas/get_stations.py
+++ b/src/indicators/flooding/glofas/get_stations.py
@@ -1,0 +1,38 @@
+"""
+Script to turn csv of GloFAS stations into config format
+for aa toolbox. Input is country ISO2 code.
+"""
+import os
+import sys
+from pathlib import Path
+
+import pandas as pd
+
+DATA_DIR = Path(os.environ["AA_DATA_DIR"])
+FILENAME = (
+    DATA_DIR
+    / "private/exploration/glb/glofas/Qgis_World_outlet_202104_20210421.csv"
+)
+
+country_code = sys.argv[1]
+
+df = pd.read_csv(
+    FILENAME,
+    usecols=[
+        "station_id",
+        "StationName",
+        "Country code",
+        "CountryName",
+        "LisfloodX",
+        "LisfloodY",
+    ],
+)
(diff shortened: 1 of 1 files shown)


src/indicators/flooding/glofas/get_stations.py
```diff
@@ -0,0 +1,38 @@
...
+import os
+import sys
+from pathlib import Path
+
+import pandas as pd
+
+DATA_DIR = Path(os.environ["AA_DATA_DIR"])
+FILENAME = (
+    DATA_DIR
+    / "private/exploration/glb/glofas/Qgis_World_outlet_202104_20210421.csv"
```
Tinkaa, line 14: what is the source of the file? How did you get it and what does it contain? Should it be updated from time to time?

src/indicators/flooding/glofas/get_stations.py
```diff
@@ -0,0 +1,38 @@
...
+        "CountryName",
+        "LisfloodX",
+        "LisfloodY",
+    ],
+)
+df = df.loc[df["Country code"] == country_code].sort_values(by="station_id")
+
+print("glofas:")
+print("  reporting_points:")
+for _, row in df.iterrows():
```
Tinkaa, line 34: Is this the best way to output? Would it make sense to save it to a yaml instead? 
If it is quickest, I am also fine with leaving it, but maybe add somewhere that that is the output

src/indicators/flooding/glofas/get_stations.py
```diff
@@ -0,0 +1,38 @@
...
+
+import pandas as pd
+
+DATA_DIR = Path(os.environ["AA_DATA_DIR"])
+FILENAME = (
+    DATA_DIR
+    / "private/exploration/glb/glofas/Qgis_World_outlet_202104_20210421.csv"
+)
+
+country_code = sys.argv[1]
```
Tinkaa, line 17: Would be more user friendly if you can also input lower case. But get that this is a quick script and not toolbox script

Tinkaa (approved):
Hey! It looks fine and works. I think if you want smth like this in toolbox as well it could use some more explanation etc, but for here it is fine. The only big comment is to add something on where you got the file from.

PR #304, opened 2022-08-05
Review of Malawi Threshold for Jan Feb

PaulineNimo:


analyses/mwi/dryspells_trigger/16_review_jan_feb_pred_threshold.md
```diff
@@ -0,0 +1,494 @@
...
+        df_pr.loc[m,["month_miss_rate"]]= miss_rate(fn,tp,epsilon)
+        df_pr.loc[m,["month_false_alarm_rate"]]= false_alarm_rate(fp,tp,epsilon)
+        df_pr.loc[m,["tn","tp","fp","fn"]]=tn,tp,fp,fn
+    df_pr=df_pr.reset_index()
+    return df_pr
+```
+
+
+```python
+def compute_confusionmatrix_leadtime(df,target_var,predict_var, ylabel,xlabel,colp_num=3,title=None,figsize=(15,8)):
```
Tinkaa, line 133: It seems this function and the next two are not used so you could delete them. But if you think you might use them in further exploration, you can leave them ofc.

analyses/mwi/dryspells_trigger/16_review_jan_feb_pred_threshold.md
```diff
@@ -0,0 +1,494 @@
...
+
+
+```python
+#define areas, months, years, and leadtimes of interest for the trigger
+sel_adm=["Southern"]
+sel_months=[1,2]
+seas_years=range(start_year,end_year)
+sel_leadtime=[3]
+
+adm_str="".join([a.lower() for a in sel_adm])
```
Tinkaa, line 299: Large parts from here till line 361 are repeated again for the jan and feb separate computations. So I would create a function where those repetititions are included. As the repetition is quite error-prone (though there doesn't seem to be any errors atm)

Tinkaa (commented):
Looks good and everything runs on my maching! Great you were able to replicate :) 
You can functionalize a few things and remove unused functions/imports. But since it is probably still quite exploratory, just find the balance that works best for you. 

Lets discuss when you are back from a methodological perspective what the next steps in determining the thresholds could be :)

Tinkaa:
@PaulineNimo I know you are very busy but let me know if you need anything else reviewed from me in the coming 2 weeks! (preferably not on the last day ;) )

PR #305, opened 2022-08-17
update readme and last year

Tinkaa:
I reran the `scr/mwi/compute_trigger_ecmwf.py` script with new data. It all worked except one bug that the `max_year` was set to 2021, so fixed that. 

Also added more documentation in that script and the readme such that it is hopefully a bit easier to understand and find what is needed. 

If you have time, it would be great if you could also try to run the `scr/mwi/compute_trigger_ecmwf.py` script to make sure it also works on your setup!


Tinkaa:
@PaulineNimo would be great if you got the time to have a look at this (and run the script) before our call Friday :)

PR #309, opened 2022-09-06
probabilities threshold iri global

Tinkaa:
Computed global return periods of reaching the 42.5% threshold on IRI forecasts. 
Return period seems very high (i.e. not common) so would be great if you could have a look if you think this approach makes sense or I am overlooking something.

### DIFF
--- a/analyses/tcd/notebooks/forecasts/tcd_iri_seas_forecast_probability.md
+++ b/analyses/tcd/notebooks/forecasts/tcd_iri_seas_forecast_probability.md
@@ -1,0 +1,154 @@
+# IRI forecast and frequency of a probability higher than 42.5%
+For TCD the trigger was set at a probability of 42.5% of below average rainfall.
+However, we only have IRI forecasts since 2017 and during those years this trigger was never met. 
+Simeltaneously no large drought occured during those years. We can therefore not compute representable performance metrics. 
+
+Instead of focusing on our region of interest, this notebooks investigates how often a probability of 42.5% has occurred globally since 2017. 
+While far from perfect, this could potentially serve as an estimate for a return period. 
+
+
+#### Load libraries and set global constants
+
+```python
+%load_ext autoreload
+%autoreload 2
+%load_ext jupyter_black
+```
+
+```python
+import pandas as pd
+import rioxarray
+import numpy as np
+import xarray as xr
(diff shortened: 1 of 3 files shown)


Tinkaa:
Actually one thing might be that blobs of raster cellls with less extreme probabilities might generally be larger and are thus dominating this analysis when looking at per raster cell? 🤔

analyses/tcd/notebooks/forecasts/tcd_iri_seas_forecast_probability.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+We load all IRI data and plot the distribution of below average probabilities. 
+We see that the median lies around 34% and the values are not normally distributed. 
+We don't see large differences in the distributions between leadtimes
+
+```python
+# C indicates the tercile (below-average, normal, or above-average).
+# F indicates the publication month, and L the leadtime
+ds_iri = get_iri_data(config, download=False)
```
caldwellst, line 51: Where does `config` come from?

caldwellst (commented):
Hey hey. Just couldn't figure out exactly what the `config` that is required is.

I tried using the config from here:

```python
from src.indicators.drought.config import Config
```

Once I get that I can check the rest of it!

caldwellst:
Ha, step to use config wasn't that hard!

What a tough problem. Some questions I have, also based on what you've asked:

- Should we restrict to similar longitudes, and maybe even just look at the Sahel for instance?
- To answer the question about hitting the area thresholds, could we look at whenever a cell hits the threshold, percent of other areas that have? Or even expand that to a look at how many cells in a radius of X hit the threshold? Basically testing if it's often that 42.5 is the highest probability we see, and we are only capturing peaks in probability that are not met in neighbouring cells that fall below it, or if the peaks are much higher? Judging by the histogram, I hope it's the latter, but unclear.
- On top of that, if we group them together, we could classify these periods and cells as 1 event, which might make it very clear how low the return period is. It could be that the 5% above the threshold actually are just a large number of cells in a very few number of years/geographic areas.

caldwellst:
Sorry I don't have time to work more on this! Thanks for your work and hopefully these comments stay relevant for us when we work on this after you leave!

caldwellst (approved):

PR #310, opened 2022-09-07
Explore uncertainty for biomasse analysis in Chad

caldwellst:
Interested to see what everybody thinks!


Tinkaa:
Hihi, sorry I am so late to the game. But I read the html in detail and I just love it! And I think it makes sense! Only thought I was having if it makes sense with any kind of distribution to look at the standard deviation. E.g. if we are interested in extremes and the distribution has a long tail. Then it might actually be real good while the std can be relatively large which might make it seem like the uncertainty is higher than it is? But tbh think this is a case we won't see often so yes definitely in favor of using it. 

Code wise, I ran the rmd and all works

PR #313, opened 2022-09-09
NER bootstrap extension

turnerm:
Oops it looks like I messed up the file rename.

But anyway I changed things so much I thought it's cleaner just to make a new notebook and save the old one for posterity (for now).

Main changes are speeding up the bootstrap and adding another population.


Tinkaa (approved):
I am gonna miss your coding proness!
I understood all of it and all of it runs so good to go imo. 
I would just suggest to remove the old notebook but I can also do that in the old branch

PR #318, opened 2022-10-19
Ner trigger report

joseepoirier:
This version of the NER trigger report was circulated with HFSA and CERF ahead of a deep dive on it 21 Oct. 

@caldwellst Do you think it is more efficient to do a proper PR on the _template_ update once that's done? Reminder: all metrics in the report are computed elsewhere where the code is/was peer-reviewed. This PR is only for the Rmd.


caldwellst:
Hey Josée! I think a code review can be best placed after the template is ready as you say. However, I exported the PDF and put some comments into it as a general document review that I hope is helpful as we work to finalise the template.

[AA Trigger Report_SC.pdf](https://github.com/OCHA-DAP/pa-anticipatory-action/files/9838532/AA.Trigger.Report_SC.pdf)

joseepoirier:
Good idea @caldwellst . Thanks for the comment, will incorporate into the template update. Thanks!

PR #320, opened 2022-10-26
Trigger report template update

joseepoirier:
Per our discussion, here is the updated template. I suspect you'll do a thorough review of the template code as you try to create the TCD report. Happy to walk you through it and/or discuss any changes you think should be made. I hit the limit of changes I could make without more clarity on what is country-agnostic vs country-specific. Thanks!


caldwellst:
@joseepoirier Based on feedback from Leonardo on the Chad trigger card and filling out the card template myself, the only changes at a template level are:

1. Changing the name from Expected Cost to Expected Annual Cost to make it more explicit it is calculated as an annual expectation.
2. I added the in a Trigger section to the table to just quickly describe what the trigger is at each timepoint.

The rest were specific changes for the Chad framework, I think. You can see the template at #321.

analyses/country_template/docs/trigger_report_template.Rmd
```diff
@@ -29,305 +29,502 @@ editor_options:
...
 knitr::opts_chunk$set(message = FALSE) # do not print messages by default
 knitr::opts_chunk$set(warning = FALSE) # do not print warnings by default
-
 options(scipen = 999) # turn off scientific notation
 options(encoding = "UTF-8") # set encoding to UTF-8 instead of ANSI
 ```
 
-```{r libraries, include = FALSE}
+```{r libraries, include = FALSE, warning = FALSE}
 packages <- c('tidyverse', 'ggthemes', 'kableExtra', 'knitr', 'flextable')
```
joseepoirier: I've started using `pacman` to handle packages. Do you want to use it here?

joseepoirier (commented):
That was a welcome fresh pass at the code! It needed cleaning, restructuring and adjustments to fit multiple use cases. 

I'm not clear on what you mean here?
>I added the in a Trigger section to the table to just quickly describe what the trigger is at each timepoint.

analyses/country_template/docs/trigger_report_template.Rmd
caldwellst: Yeah, here you could use `pacman`!

caldwellst:
Sorry, to be clear, in the Trigger Mechanism Snapshot portion of the template for Chad, I added a `Trigger` section that actually describes what the threshold is. It is currently not there and thought that's the only missing piece that would be useful to include in that table, so adding here in case you wanted to put into the template.

joseepoirier:
> Sorry, to be clear, in the Trigger Mechanism Snapshot portion of the template for Chad, I added a `Trigger` section that actually describes what the threshold is. It is currently not there and thought that's the only missing piece that would be useful to include in that table, so adding here in case you wanted to put into the template.

Ha, I see. Yes I think the current introductory paragraph is quite meager. I don't see the paragraph in the TCD report.  Feel free to introduce the change in the template or copy/paste it here and I will.

joseepoirier:
> Sorry, to be clear, in the Trigger Mechanism Snapshot portion of the template for Chad, I added a `Trigger` section that actually describes what the threshold is. It is currently not there and thought that's the only missing piece that would be useful to include in that table, so adding here in case you wanted to put into the template.

Finally reviewed the code carefully. I see the addition of a line in the Snapshot table (from your comment here I expected a paragraph.) Good addition. I recommended renaming it to "Threshold" rather than "Trigger".

PR #321, opened 2022-12-07
Tcd metrics

caldwellst:
Implemented the Chad trigger card based on the discussions we've had regarding the rule of three and bootstrapping.


caldwellst:
Leonardo's comments and then my responses, from Slack:

> Thanks! I am surprised that the activation for trigger 1 and 2 is quite low. Have you taken into account the fact that we have 2 trigger points instead of just one? We have seen that this has a significant impact as the forecasts for different months are not necessarily highly correlated

Yes, this accounts for the multiple triggers. Individually they have about 5% chance of reaching that threshold.

> Shouldn’t the target period always be Jul-Sep for trigger 1 and 2?

Yes, typo! Will address.

> Can you add somewhere a description (maybe under Key messages - Shock detection) of how the actual shocks are estimated?

This is added beneath the all shocks graph under notes. This is where we listed in the Niger report.

> What is the expected cost in the table? Is it the expected annual cost on average? If that’s the case let’s make it explicit

The expected cost is again following the Niger template, where it’s the central likelihood of activation multiplied by the cost. How would you recommend making it explicit? (Leonardo said yes let's make it explicit)

caldwellst:
I added just an asterisk next to Expected Cost to indicate it's covered in the notes, but I'm sure there's a better solution.

caldwellst:
Have updated to call it "Expected Annual Cost" baed on feedback from Leonardo.

joseepoirier:
Sorry Seth I don't see an email notifying me of this PR. Will review shortly. 

Two things I already noticed: 
(1) there is no learning table, which there should be even if you don't report the full set of metrics that we had for NER
(2) plots: i believe I found a way to keep the plots on Gdrive `Data`. I'll test it on the template and let you know if/when we can move them.

caldwellst:
What do you mean by learning table? The Learning Opportunities table is still there in the report at the bottom, but let me know if it's something else I missed. I took the Niger report and directly changed that so hopefully nothing missing.

joseepoirier:
> What do you mean by learning table? The Learning Opportunities table is still there in the report at the bottom, but let me know if it's something else I missed. I took the Niger report and directly changed that so hopefully nothing missing.

Ha no, it's there indeed. I had forgotten it was stored on `Data`. I'll take a closer look at this PR asap. Sorry it's confusing this is my third of fourth version of a report in 2 days! 😂

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+mitigation_warranted <- ifelse(t3_far_low_end_68 >= 40, "Mitigation is recommended to minimise the negative impact of a false alarm due to its moderate to high likelihood.", "")
+
+```
+
+# Anticipatory Action Trigger Mechanism Report
+<p style="color:#007ce1;font-size:22px;text-align:center;"> `r paste0(shock, ' in ', country_name)`</p> 
+<p style="color:black;font-size:14px;text-align:right;"> Last updated: `r Sys.Date()`</p> 
+
+## Key Messages
+* <span style="color: blue;">LIMITED HISTORICAL DATA</span> <p style="color:black;font-size:16px;text-align:left;"> Triggers 1 and 2 rely on the IRI seasonal forecast, with limited data from 2018 to today. Likelihood of activation estimated based on global data, for Triggers 1 and 2 together, but not together with Trigger 3, and no performance metrics are calculated.</p>
```
joseepoirier: I thought historical forecasts went back to 2017?

joseepoirier: Unpack the second sentence. What is "global" referring to? WHy split T1+T2 from T3? Why no performance metrics? What does it mean for our understanding of likelihood and risk in disbursing funds based on this mechanism? RIght now it reads more like a footnote than an important warning on limitations.

Also this is the key messages section. Consider breaking it up the first point and/or develop to provide more context later in the doc.

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+* <span style="color: blue;">ACTIVATION</span> <p style="color:black;font-size:16px;text-align:left;"> There is a `r paste0(t3_atv_low_end_68, ' - ', t3_atv_high_end_68)`% chance that Trigger 3 will activate in a given year. Globally, thresholds are met for Trigger 1 or 2 `r t12_atv_central`% of the time, however there is high uncertainty on how frequently this would occur in `r country_name`.</p>
```
joseepoirier: `Globally` again won't be understood by readers.

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+                                     monitored_area_a,
+                                     timepoints_a, 
+                                     targeted_window_a, 
+                                     lead_time_a, 
+                                     data_source_a,
+                                     trigger_a,
+                                     hist_period_a,
+                                     monitored_by_a,
+                                     package_a, 
+                                     paste0(prob_a, "%, based on global IRI data, but very high uncertainty"),
```
joseepoirier, line 343: The pilot team & HQ likely expect something around 35%. This is particularly low. Perhaps worth a separate footnote to elaborate?

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+            , color = '#1bb580') %>%
+  kableExtra::footnote(general_title = "Notes",
+                       general = trigger_table_footnote)
+
+```
+
+&nbsp;
+
+## Activation Scenarios
+<p style="color:black;font-size:16px;text-align:left;"> Due to the lack of historical data for the IRI seasonal forecast, we do not assess the likelihood of activation scenarios together. Refer to the likelihood of the individual triggers activating.</p>
```
joseepoirier: It is unclear to the reader why scenarios cannot be estimated if `atv` is computed for individual triggers.

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+
+&nbsp;
+
+## Technical Team
+<p style="color:black;font-size:16px;text-align:left;"> `r data_providers` have provided data and/or analytical support. A total of `r nbr_hist_shocks` historical occurrences of severe shocks were documented and used to develop the trigger mechanism. `r analysis` </p>
```
joseepoirier: There were 5 historical shocks / years documented, but only 5 or so years of forecast to work with. how many shocks occurred on years for which we had forecasts? What time span do we have historical shock data for?  

Important to be clear that we could only set a threshold and estimate performance based on 5 years of data for which we had forecasts + impact data.

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+# trigger A info  
+name_a <- "Trigger1" # what name is the trigger listed as in the perf_metrics_table.csv files?
+type_a <- 'Predictive' # Predictive or Observational?
+monitored_area_a <- "Central Chad"
+timepoints_a<- 'Mar, Apr' # eg: 'Jan, Feb, Mar'
+package_a <- 'Package 1' # eg: 'Package 1'
+funding_a <- 3.71 # input number only, no $ sign or 'm' for million. Formatting will be done automatically.
+targeted_window_a <- 'Jul-Sep' # 3-letter names for first and last month. eg: 'Jul-Sep'. Forecasted period or period over which obs data are computed.
+lead_time_a <- '4-3 months' # 'eg: 7-5 months', Change units as needed. Include all lead times if multiple activation timepoints
+data_source_a <- 'IRI seasonal forecast' # name of product + provider
```
joseepoirier: add "tercile" to data sources (we also use their next gen / flexible forecasts)

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+trigger_a <- '42.5% below average probability'
```
joseepoirier: rename `threshold_a` (and b and c) for disambiguation

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+trigger_list <- c(name_c) # limited due to issues with historical data in Triggers 1 and 2
+scenario_list <- c() # No scenarios given limits of data
+
+# Table footnotes
+trigger_table_footnote <- "Expected annual cost computed using most likely probability of activation (central value). All triggers can reach their threshold and activate independently from one another. Triggers 1 and 2 did not have sufficient data for testing performance on Chad, but likelihood of activation tested on global data. Trigger 3 performance assessed by comparing recommended activations with historical bad years (i.e., years with a shock). 'Bad years' were identified through convergence of evidence reviewed from historical CERF allocations, FEWS NET reports, EM-DAT data, World Bank statistics and the 2015 and 2017 multi-risk contingency plans for Chad."
+
+full_atv_definition <- "A full activation occurs when all three triggers met and all packages are disbursed."
+package1_atv_definition <- "Package 1 is activated when Trigger 1 is met."
+package2_atv_definition <- "Package 2 is activated when Trigger 2 is met."
+package3_atv_definition <- "Package 3 is activated when Trigger 3 is met."
```
joseepoirier, line 104: The table lists Trigger 3 with Package 2. Is there a Package 3?

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+&nbsp;
+```{r mechanism-snapshot, include = TRUE}
+
+snapshot_df <- data.frame(row_name = c('Type', 
+                                       'Monitored Area',
+                                       'Activation Timepoints', 
+                                       'Target Period', 
+                                       'Lead Time', 
+                                       'Data Source',
+                                       'Trigger',
```
joseepoirier: `Threshold` would be less ambiguous and more informative

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+                                       'Expected Annual Cost',
+                                       'All Activations (%)', 
+                                       'All Shocks (%)'),
+                       trigger_a = c(type_a, 
+                                     monitored_area_a,
+                                     timepoints_a, 
+                                     targeted_window_a, 
+                                     lead_time_a, 
+                                     data_source_a,
+                                     trigger_a,
```
joseepoirier: `threshold_a`

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+                                     expected_cost_a,
+                                     "N/A", 
+                                     "N/A"),
+                       trigger_b = c(type_b,
+                                     monitored_area_b,
+                                     timepoints_b, 
+                                     targeted_window_b, 
+                                     lead_time_b, 
+                                     data_source_b,
+                                     trigger_b,
```
joseepoirier: `threshold_b`

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+                                     expected_cost_b,
+                                     "N/A",
+                                     "N/A"),
+                       trigger_c = c(type_c, 
+                                     monitored_area_c,
+                                     timepoints_c, 
+                                     targeted_window_c, 
+                                     lead_time_c, 
+                                     data_source_c,
+                                     trigger_c,
```
joseepoirier: `threshold_c`

analyses/tcd/metrics/TCD_trigger_report.Rmd
```diff
@@ -0,0 +1,432 @@
...
+trigger_b <- '42.5% below average probability'
+hist_period_b <- "2018 - Present"
+monitored_by_b <- 'CHD'
+  
+# trigger C info  
+name_c <- "Trigger3"
+type_c <- 'Observational'
+monitored_area_c <- "Central Chad"
+timepoints_c <- 'Sep'
+package_c <- 'Package 2'
```
joseepoirier: Isn't this Package 3?

analyses/tcd/metrics/tcd_metrics.R
```diff
@@ -0,0 +1,160 @@
...
+  biomasse_pred = factor(biomasse_anomaly <= 80, levels = c(T,F)),
+  bm_end_season = factor(year %in% biomasse_years, levels = c(T,F)),
+  drought_list = factor(year %in% list_years, levels = c(T,F))
+)
+
+#################
+#### METRICS ####
+#################
+
+# generate metrics for months July to December
```
joseepoirier: Interesting new case for the trigger report: how do we compute and communicate performance for observational triggers that can activate on a frequent basis? In Malawi, that would be daily. (In Niger that was only once.)

In this case because it's once monthly, we're treating it like a predictive trigger. Maybe there is a parallel with sudden-onset triggers here.

Just food for thought for now.

analyses/tcd/metrics/tcd_metrics.R
```diff
@@ -0,0 +1,160 @@
...
+      mis = 1 - det,
+      acc = accuracy_vec({{ true }}, {{ pred }}),
+      atv = sum({{ pred }} == TRUE) / n()
+    )
+}
+
+pred_sep_df <- pred_df %>%
+  mutate(month_pub = month + 1) %>%
+  filter(
+    day == 11,
```
joseepoirier, line 76: Should this be >= in case the data is published after the 11th?

analyses/tcd/metrics/tcd_metrics.R
```diff
@@ -0,0 +1,160 @@
...
+    metric, trigger, value, point
+  ) 
+
+# Saving in Trigger 1 and 2 metrics
+# Calculated in `tcd_iri_seas_forecast_probability.md`
+
+df_t12 <- data.frame(
+  metric = "atv",
+  trigger = c("Trigger1", "Trigger2", "Trigger1-2"),
+  value = c(0.070746, 0.079827, 0.111824),
```
joseepoirier, line 120: Is there a programmatic way to get these values?
Why no `low_end` and `high_end` values?

analyses/tcd/notebooks/forecasts/tcd_iri_seas_forecast_probability.md
```diff
@@ -121,8 +121,8 @@ df_bavg["meet_thresh"] = np.where(df_bavg.prob >= thresh, True, False)
 ```
 
 ```python
-#get sum of leadtimes that predict above the threshold
-#for the given predicted period and coordinates
+# get sum of leadtimes that predict above the threshold
```
joseepoirier, line 124: I don't follow why you sum lead times?

analyses/tcd/notebooks/forecasts/tcd_iri_seas_forecast_probability.md
```diff
@@ -121,8 +121,8 @@ df_bavg["meet_thresh"] = np.where(df_bavg.prob >= thresh, True, False)
...
+# for the given predicted period and coordinates
```
joseepoirier, line 125: Explain why you are using coordinates, and why those specifically.

analyses/tcd/notebooks/forecasts/tcd_iri_seas_forecast_probability.md
```diff
@@ -121,8 +121,8 @@ df_bavg["meet_thresh"] = np.where(df_bavg.prob >= thresh, True, False)
 ```
```
joseepoirier, line 121: Add note on how the threshold was selected / reference to another doc

joseepoirier (commented):
Hey Seth, this looks close to ready to go. Main recommendations concern (1) adding more information so every variable/decision is explained even if only to refer to other scripts and (2) to strengthen the narration so decision-makers reading the report get the point more clearly that there is a lot of unknown with this trigger mechanism, and why. Then, they'll be able to deliberate whether or not they believe the trigger is justifiable to disburse funds, or know what questions to ask to understand those limitations.

I saw in your "thoughts" section that you think we might be doing something wrong. I assume by now this has been discussed and talked through with Leonardo and has been resolved. If so, the section would benefit from an update to reflect that.

analyses/tcd/metrics/TCD_trigger_report.Rmd
caldwellst: Historical forecasts didn't include the periods for 2017 that we are interested in, so for this pilot, data was only available until 2017.

Have made this read more like a key point and will add more context down below in the document.

analyses/tcd/metrics/TCD_trigger_report.Rmd
caldwellst: Have clarified in the above and slightly edited this portion so should be clear now.

analyses/tcd/metrics/TCD_trigger_report.Rmd
caldwellst: Yep, good catch! We have three separate packages.

analyses/tcd/metrics/tcd_metrics.R
caldwellst: In this case, we actually only look at it in September, not earlier, but were calculating the performance to see when we might want to set our activation timepoints. That was just a comment that accidentally was left in the code, so actually the same as Niger here. Interesting thought, though. I think we could treat it like a predictive trigger if we are doing something like monthly (so as in Niger, you could look at the individual trigger points). However, clearly impossible for daily triggers. Some interested ideas could be there for plots maybe of likelihoods across the observation period, maybe?

analyses/tcd/metrics/tcd_metrics.R
caldwellst: It is alright as is as this is static, and also was just to produce this data (not monitor). The day is set based on the dekadal definitions, so is always the 11th.

analyses/tcd/metrics/tcd_metrics.R
caldwellst: There was no way to get a confidence interval that we were comfortable with because the bounds were much too small since it was estimating our global intervals. So we just wanted to put together the values for these specific ones as our rough estimates.

analyses/tcd/notebooks/forecasts/tcd_iri_seas_forecast_probability.md
caldwellst: It's to calculate the percentiles, getting total number within each lead-time that met threshold, then dividing by total number.

caldwellst:
Thanks Josée! Have gone through and addressed. Not sure what you mean about me thinking we did something wrong? Would you mind pointing me to this? I thought might be in the Learning Opportunities section but couldn't figure out exactly what that might be.

I like the thought about how to report on observational triggers that are frequently observed. Actually not the case here, we only observe once like in Niger, but interesting stuff about Malawi. I made an issue about it so we don't lose the thought when this PR is closed!

joseepoirier:
Hey Seth, 

A doc/line reference would have been helpful, sorry. In `tcd_iri_seas_forecast_probability.md`, now line 172. You write "I am therefore wondering if we are doing this correctly.." 

It would be good for you, @turnerm and I to brainstorm performance metrics for continuous triggers. Monica's already done a lot of thinking on this for sudden-onset triggers and her input would be valuable as we figure out the case of observational triggers.

Lastly, is there an explanation for why the forecasts were global and not "cropped" to Chad? If not, it seems relevant to include.

caldwellst:
Aaah I see. That's very old code. All of that was actually Tinka originally, I was just making a few amendments to the file to get it running properly on my machine. And things like that comment are also from her. And agreed on the discussion and brainstorming, I think will be most useful to do that when we have a specific case to work on, like Malawi. Have clarified why using global forecasts for the likelihood of activation!

PR #322, opened 2022-12-13
Update ARC2 code for trigger monitoring

caldwellst:
Updated the code for ARC2 monitoring of the Malawi trigger. It is fully functioning on my machine now. Can you see if you can also get it to run on yours tomorrow? You should be able to see December 13th data in the aggregated and raw data after running. Hope that this functions well.


PaulineNimo:
Thanks, it is running well now. The only thing is in the second last cell, because of the empty data frame (possibly due to not being within the monitoring period), there is an error with generating the plot but this seems okay now.

caldwellst:
Ah great! Yeah, you can test by changing the monitoring period to something earlier if you want. But otherwise, I think that means we good to go!

